PR #29, opened 2021-11-03
Workflow

orianac:


orianac:
Thanks in advance for code reviews @tcchiao , @norlandrhagen , @jhamman !! 🍰 🥇 🍩 🧁 
There are a few big to-dos I'm noting here for future reference:
* It'd be good to figure out how to draw the hyperparameters from the context as opposed to passing them via a dictionary in the way it's going now. Any ideas on that would be much appreciated! Sounds like @jhamman might have been playing with prefect recently so I'm all ears!
* Clean up routines would be great to implement. There are a bunch of temporary files that get written (currently) to `az://cmip6/temp/`. These can definitely get wiped. There are others that get written to `az://cmip6/intermediates/` that we might want to save. I defer to @norlandrhagen and @jhamman on the ideas about the organization there.
* I haven't actually shipped the `bcsd_flow.py` to prefect cloud, but excited to try it out!!

cmip6_downscaling/workflows/utils.py
```diff
@@ -25,12 +27,67 @@ def get_store(prefix, account_key=None):
...
+    """
+
+    coarse_obs_path = f"{workdir}/intermediates/{OBS}_{GCM}_{TRAIN_PERIOD_START}_{TRAIN_PERIOD_END}_{VARIABLE}.zarr"
+    spatial_anomolies_path = f"{workdir}/intermediates/anomalies_{OBS}_{GCM}_{TRAIN_PERIOD_START}_{TRAIN_PERIOD_END}_{VARIABLE}.zarr"
+    bias_corrected_path = f"{workdir}/intermediates/bc_{OBS}_{SCENARIO}_{GCM}_{TRAIN_PERIOD_START}_{TRAIN_PERIOD_END}_{VARIABLE}.zarr"
+    final_out_path = f"{outdir}/bcsd_{OBS}_{SCENARIO}_{GCM}_{PREDICT_PERIOD_START}_{PREDICT_PERIOD_END}_{VARIABLE}.zarr"
+    return coarse_obs_path, spatial_anomolies_path, bias_corrected_path, final_out_path
+
+
+def load_paths(paths):  # What type do i use here since paths is of unknown length? : list[str]):
```
tcchiao, line 80: `List[str]`. this function might also be easier to use with `paths: Union[str, List[str]]` then return ds if `isinstance(paths, list)`, etc

cmip6_downscaling/workflows/utils.py
```diff
@@ -39,42 +96,92 @@ def delete_chunks_encoding(ds):
...
+        Stores where rechunker will write and the path to the target store
+    """
+    path_tmp = "az://cmip6/temp/{}.zarr".format(temp_file_name())
+    path_tgt = "az://cmip6/temp/{}.zarr".format(temp_file_name())
+    temp_store = fsspec.get_mapper(path_tmp, connection_string=connection_string)
+    target_store = fsspec.get_mapper(path_tgt, connection_string=connection_string)
+    return temp_store, target_store, path_tgt
+
+
+def rechunk_zarr_array(
```
tcchiao, line 121: wish list for this function :) 
1. make it work for all variables in the ds, 
2. make it "lazy" so one can call it just in case without incurring unnecessary overhead.

cmip6_downscaling/workflows/bcsd_flow.py
```diff
@@ -0,0 +1,112 @@
...
+        )
+    ),
+    adapt_kwargs={"minimum": 4, "maximum": 20},
+)
+
+
+flow_name = run_hyperparameters.pop("FLOW_NAME")  # pop it out because if you leave it in the dict
+# but don't call it as a parameter it'll complain
+
+make_flow_paths_task = task(make_flow_paths, log_stdout=True, nout=4)
```
norlandrhagen, line 56: What is the advantage of making the imported functions task here vs in cmip6_downscaling.workflows.utils?

No idea what the preferred way, just curious on your thinking!

cmip6_downscaling/workflows/bcsd_flow.py
```diff
@@ -0,0 +1,112 @@
...
+    "SCENARIO": "ssp370",
+    "TRAIN_PERIOD_START": "1990",
+    "TRAIN_PERIOD_END": "1990",
+    "PREDICT_PERIOD_START": "2090",
+    "PREDICT_PERIOD_END": "2090",
+    "VARIABLE": "tasmax",
+    "OBS": "ERA5",
+}
+
+storage = Azure("prefect")
```
norlandrhagen, line 31: Maybe we could move out the config stuff into some shared file and import it?
Maybe this is a future PR..

cmip6_downscaling/methods/bcsd.py
```diff
@@ -0,0 +1,249 @@
...
+    Returns
+    -------
+    y_rechunked_path, X_train_rechunked_path, X_predict_rechunked_path : str, str, str
+        Paths to to write y, X_train, and X_predict ready for eventual use by model
+    """
+    # load in coarse obs as xarray ds
+    [y] = load_paths([coarse_obs_path])
+    # TODO: add xarray_schema test here for chunking - if the chunking fails then try rechunking
+    # if passes skip
+    y_rechunked, y_rechunked_path = rechunk_zarr_array(
```
norlandrhagen, line 129: nit: It seems like y_rechunked and x_train_rechunked (line 142) are both unused.

cmip6_downscaling/methods/bcsd.py
```diff
@@ -0,0 +1,249 @@
...
+    -------
+    final_out_path : str
+        location where final BCSD data was written
+    """
+    y_predict, spatial_anomalies = load_paths([y_predict_path, spatial_anomalies_path])
+
+    # TODO: test - create sample input, run it through rechunk/regridder and then
+    # assert that it looks like i want it to
+    y_predict_fine = regrid_dataset(y_predict, spatial_anomalies, variable, connection_string)
+    bcsd_results = y_predict_fine.groupby("time.month") + spatial_anomalies
```
norlandrhagen, line 246: nit/question: Will "time.month" change in future runs? Should this be an arg or is it fine as is.

norlandrhagen (commented):
Looks awesome @orianac! I really like the refactor of logic, the detailed docstrings etc. Huge improvement over the last iteration.

cmip6_downscaling/data/observations.py
```diff
@@ -34,112 +32,76 @@ def open_era5(var):
...
     Parameters
     ----------
-    coarse_obs : [type]
-        [chunked in space (lat=-1,lon=-1,time=1)]
-    fine_obs : [type]
-        [chunked in space (lat=-1,lon=-1,time=1)]
+    coarse_obs : xr.Dataset
+        Coarsened to a GCM resolution. Chunked along time.
+    fine_obs : xr.Dataset
+        Original observationa spatial resolution. Chunked along time.
```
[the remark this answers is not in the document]
orianac: now implemented!

cmip6_downscaling/data/cmip.py
```diff
@@ -163,38 +163,38 @@ def cmip():
...
-        [grid_labels in CMIP6 catalog], by default ["gn"]
+        grid_labels in CMIP6 catalog, by default ["gn"]
     variable_ids : list, optional
-        [variable_ids in CMIP6 catalog], by default ['tasmax']
+        variable_ids in CMIP6 catalog, by default ['tasmax']
 
     Returns
     -------
-    [dictionary]
+    ds : xr.Dataset
```
jhamman, line 197: This function also returns a zarr group, depending on the value of `return_type`

cmip6_downscaling/data/cmip.py
```diff
@@ -163,38 +163,38 @@ def cmip():
...
-    """Loads CMIP6 GCM dataset dictionary based on input criteria.
+def load_cmip(
+    activity_ids: list = ["CMIP"],
+    experiment_ids: list = ["historical"],
+    member_ids: list = ["r1i1p1f1"],
+    source_ids: list = ["MIROC6"],
+    table_ids: list = ["day"],
+    grid_labels: list = ["gn"],
+    variable_ids: list = ["tasmax"],
+    return_type: list = 'zarr',
```
jhamman, line 174:
```suggestion
    return_type: list = 'str',
```

cmip6_downscaling/data/cmip.py
```diff
@@ -163,38 +163,38 @@ def cmip():
...
-):
-    """Loads CMIP6 GCM dataset dictionary based on input criteria.
+def load_cmip(
+    activity_ids: list = ["CMIP"],
+    experiment_ids: list = ["historical"],
+    member_ids: list = ["r1i1p1f1"],
+    source_ids: list = ["MIROC6"],
+    table_ids: list = ["day"],
+    grid_labels: list = ["gn"],
+    variable_ids: list = ["tasmax"],
```
jhamman, line 173:
```suggestion
    activity_ids: list[str] = ["CMIP"],
    experiment_ids: list[str] = ["historical"],
    member_ids: list[str] = ["r1i1p1f1"],
    source_ids: list[str] = ["MIROC6"],
    table_ids: list[str] = ["day"],
    grid_labels: list[str] = ["gn"],
    variable_ids: list[str] = ["tasmax"],
```

cmip6_downscaling/data/cmip.py
```diff
@@ -220,19 +220,48 @@ def load_cmip_dictionary(
     elif return_type == 'xr':
         ds = xr.open_zarr(stores[0], consolidated=True)
 
+    # flip the lats if necessary and drop the extra dims/vars like bnds
+
```
jhamman, line 224:
```suggestion
    # flip the lats if necessary and drop the extra dims/vars like bnds
```

cmip6_downscaling/data/cmip.py
```diff
@@ -220,19 +220,48 @@ def load_cmip_dictionary(
...
     ds = gcm_munge(ds)
 
     return ds
 
 
-def convert_to_360(lon):
+def convert_to_360(lon: Union[float, int]) -> Union[float, int]:
```
jhamman, line 230: typing comments:
- Do we expect to ever pass integers in here?
- I suspect mypy will complain when you pass a array-like type in here

cmip6_downscaling/data/observations.py
```diff
@@ -1,13 +1,11 @@
 import os
```
jhamman, line 1: What do you think about renaming this file to `training.py`? After all, ERA5 is not exactly _observations_.

cmip6_downscaling/methods/bcsd.py
```diff
@@ -0,0 +1,249 @@
...
+        Paths to where the outputs were written. These will be cached by prefect.
+    """
+
+    """
+
+    TODO: do we want to assign this to a class? write out? probably
+    """
+    # TODO: add in functionality to label the obs/spatial anomalies by grid name as opposed to GCM
+    #  since it will just help us not repeat the coarsening/spatial anomalies step
+    if rerun:
```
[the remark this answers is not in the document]
orianac: resolved

cmip6_downscaling/methods/bcsd.py
```diff
@@ -0,0 +1,249 @@
...
+        obs_ds = load_obs(obs_id, variable, time_period=slice(train_period_start, train_period_end))
```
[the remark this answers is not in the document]
orianac: great point!

cmip6_downscaling/data/observations.py
```diff
@@ -34,112 +32,76 @@ def open_era5(var):
...
-    obs_coarse = regridder(obs)
-    return obs_coarse
+        return open_era5(variable).sel(time=time_period)
 
 
-def get_spatial_anomolies(coarse_obs, fine_obs):
-    # check if this has been done, if do the math
-    # if it has been done, just read them in
-    """[summary]
+def get_spatial_anomolies(coarse_obs, fine_obs, variable, connection_string) -> xr.Dataset:
```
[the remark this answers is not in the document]
jhamman: I'm not sure this function belongs in `observations.py`?

cmip6_downscaling/data/observations.py
```diff
@@ -34,112 +32,76 @@ def open_era5(var):
...
+    """Calculate the seasonal cycle (12 timesteps) spatial anomaly associated
```
jhamman, line 71: For doc strings with longer descriptions (like this one), its best to start with a 1-line summary then follow with a longer paragraph description.

So something like this:

```
"""
Calculate the seasonal cycle (12 timesteps) spatial anomaly

The spatial anomaly is calculated by  aggregating the fine_obs to a given coarsened scale and 
then reinterpolating it back to the original spatial resolution. The outputs of this function are
dependent on three parameters:
    * a grid (as opposed to a specific GCM since some GCMs run on the same grid)
    * the time period which fine_obs (and by construct coarse_obs) cover
    * the variable
"""

File: cmip6_downscaling/data/observations.py [hunk 18]
```diff
@@ -34,112 +32,76 @@ def open_era5(var):
...
+    time_period : slice
+        time period you want to subset. e.g. slice('1985', '2015'). using
+        full years is recommended if you are going to be doing the spatial
+        anomaly calculations.
 
     Returns
     -------
-    [xarray dataset]
-        [Chunked {time:-1,lat=10,lon=10}]
+    open_era5(variable).sel(time=time_period) : xr.Dataset
```
jhamman, line 60: I think the [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html#returns) way to reference an unnamed returned object is:
```suggestion
    xr.Dataset
```

cmip6_downscaling/methods/bcsd.py
```diff
@@ -0,0 +1,249 @@
...
+        Path to write coarsened obs zarr dataset (e.g. 'az://bucket/out.zarr')
+    connection_string : str
+        Connection string to give you read/write access to the out buckets specified above
+    rerun : bool, optional
+        [description], by default True
+
+    Returns
+    -------
+    coarse_obs_path, spatial_anomolies_path : str, str
+        Paths to where the outputs were written. These will be cached by prefect.
```
jhamman, line 59: split this out into separate lines

cmip6_downscaling/methods/bcsd.py
```diff
@@ -0,0 +1,249 @@
...
+    obs_id: str,
+    train_period_start: str,
+    train_period_end: str,
+    variable: str,
+    coarse_obs_path: str,
+    spatial_anomolies_path: str,
+    connection_string: str,
+    rerun: bool = True,
+) -> tuple[str, str]:
+    """Given GCM and selected obs dataset for a given variable and a given time period, at the coarse scale to match that
```
jhamman, line 30: see comment above on a short 1-line description.

cmip6_downscaling/methods/bcsd.py
```diff
@@ -0,0 +1,249 @@
...
+    variable : str
+        Variable of interest in CMIP conventions (e.g. 'tasmax')
+    coarse_obs_path : str
+        Path to write coarsened obs zarr dataset (e.g. 'az://bucket/out.zarr')
+    spatial_anomolies_path : str
+        Path to write coarsened obs zarr dataset (e.g. 'az://bucket/out.zarr')
+    connection_string : str
+        Connection string to give you read/write access to the out buckets specified above
+    rerun : bool, optional
+        [description], by default True
```
jhamman, line 54: note that we can remove this once we have caching implemented at the prefect level.

cmip6_downscaling/methods/bcsd.py
```diff
@@ -0,0 +1,249 @@
...
+    predict_period_start : str
+        Date for prediction period start (e.g. '2070')
+    predict_period_end : str
+        Date for prediction period end (e.g. '2099')
+    variable : str
+        Variable of interest in CMIP conventions (e.g. 'tasmax')
+
+    Returns
+    -------
+    y_rechunked_path, X_train_rechunked_path, X_predict_rechunked_path : str, str, str
```
jhamman, line 122: split into separate lines

cmip6_downscaling/methods/bcsd.py
```diff
@@ -0,0 +1,249 @@
...
+        connection_string=connection_string,
+        max_mem='1GB',
+    )
+
+    X_predict = load_cmip(
+        source_ids=[gcm], activity_ids=['ScenarioMIP'], experiment_ids=[scenario], return_type='xr'
+    ).sel(time=slice(predict_period_start, predict_period_end))
+    assert len(X_predict.lat) == len(
+        X_train.lat
+    ), "Uh oh! Your prediction length is different from your training and so the chunk sizes wont match and xarray will complain!"
```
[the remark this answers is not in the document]
jhamman: And, if this check is needed, lets raise a `ValueError` here instead of using an assert statement.

cmip6_downscaling/methods/bcsd.py
```diff
@@ -0,0 +1,249 @@
...
+
+
+def fit_and_predict(
+    X_train_path: str,
+    y_path: str,
+    X_predict_path: str,
+    bias_corrected_path: str,
+    dim: str = "time",
+    variable: str = "tasmax",
+):
```
jhamman, line 177: type return value

cmip6_downscaling/methods/bcsd.py
```diff
@@ -0,0 +1,249 @@
...
+    return bias_corrected_path
+
+
+def postprocess_bcsd(
+    y_predict_path: str,
+    spatial_anomalies_path: str,
+    final_out_path: str,
+    variable: str,
+    connection_string: str,
+):
```
jhamman, line 219: type return value

cmip6_downscaling/workflows/bcsd_flow.py
jhamman: I think I agree here. It seems like most of these functions should move here and just be Prefect Tasks.

cmip6_downscaling/methods/bcsd.py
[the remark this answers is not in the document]
orianac: this was a work-around but i implemented a stronger fix. the pointwise models need the spatial chunks to be the same between fit and predict steps. since we were algorithmically determining a 100mb chunk size (as opposed to specifying it), if we fed a different length of temporal record to predict, it might result in a different spatial chunk size. now i allow for either allowing the chunk size calculation or for specfying it. but it's good to keep in mind that this could cause a performance issue if we predict on much longer timescale than we train, since the spatial chunksize  (and thus the entire chunksize since there is only ever one chunk in time for pointwise models) will be determined by the train period, and it could result in larger chunks. for example, if chunk size of 30 years in train were 100mb, using the same spatial chunk but extending to a 150 year predict window would result in 500mb chunks.

cmip6_downscaling/workflows/utils.py
```diff
@@ -111,3 +242,61 @@ def calc_auspicious_chunks_dict(ds, target_size='100mb', chunk_dims=('lat', 'lon
...
+    ds_regridded : xr.Dataset
+        Final regridded dataset
+    """
+    # TODO: use xarray_schema to check that the dataset is chunked into map space already
+    # and if not rechunk it into map space (only do the line below if it's necessary)
+    ds_rechunked, ds_rechunked_path = rechunk_zarr_array(
+        ds, connection_string, variable, chunk_dims=('time',), max_mem="1GB"
+    )
+    regridder = xe.Regridder(ds_rechunked, target_grid_ds, "bilinear", extrap_method="nearest_s2d")
+    regridder = xe.Regridder(ds_rechunked, target_grid_ds, "bilinear", extrap_method="nearest_s2d")
```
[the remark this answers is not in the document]
orianac: good eye!

cmip6_downscaling/workflows/utils.py
```diff
@@ -25,12 +27,67 @@ def get_store(prefix, account_key=None):
     return store
 
 
+def make_flow_paths(
```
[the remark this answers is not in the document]
orianac: good call- moved.

cmip6_downscaling/workflows/utils.py
orianac: it is lazy now!

orianac: Passing (1) to an issue.

cmip6_downscaling/workflows/bcsd_flow.py
orianac: Currently the `bcsd_flow.py` script contains a prefect flow (e.g. there is a `flow.run()` call at the bottom). If the functions were specified in there we wouldn't be able to import them anywhere else without running the flow. If you know of a way to bypass that, let me know, but I recommend keeping it as-is. Especially if you might want to restructure things, as testing any changes would likely be frustrating.

cmip6_downscaling/methods/bcsd.py
orianac: good catch! also x_predict_rechunked. i've changed them to `_`

cmip6_downscaling/methods/bcsd.py
orianac: good question! this will always be `time.month`. this method works by calculating seasonal averages (aka one value for every month of the year- 12 values). if we were to switch to daily bcsd we'd have to change a lot of other things, so i think this is fine as-is.

cmip6_downscaling/data/observations.py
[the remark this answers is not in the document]
orianac: moved it to utils

cmip6_downscaling/data/cmip.py
```diff
@@ -163,38 +163,38 @@ def cmip():
...
+        grid_labels in CMIP6 catalog, by default ["gn"]
     variable_ids : list, optional
-        [variable_ids in CMIP6 catalog], by default ['tasmax']
+        variable_ids in CMIP6 catalog, by default ['tasmax']
 
     Returns
     -------
-    [dictionary]
+    ds : xr.Dataset
         [dictionary containing available xarray datasets]
```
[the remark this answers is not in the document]
orianac: i'm not accepting this but since i am changing the text, but will remove the brackets

cmip6_downscaling/data/cmip.py
orianac: i think you meant to make this `str = 'zarr'`, which is great! at least that's what i'm implementing! let me know if otherwise.

cmip6_downscaling/methods/bcsd.py
orianac: done

cmip6_downscaling/methods/bcsd.py
orianac: done

cmip6_downscaling/methods/bcsd.py
orianac: done

cmip6_downscaling/methods/bcsd.py
```diff
@@ -0,0 +1,249 @@
...
+    # TODO: add in functionality to label the obs/spatial anomalies by grid name as opposed to GCM
+    #  since it will just help us not repeat the coarsening/spatial anomalies step
+    if rerun:
+        obs_ds = load_obs(obs_id, variable, time_period=slice(train_period_start, train_period_end))
+
+        gcm_one_slice = load_cmip(return_type='xr', variable_ids=[variable]).isel(time=0)
+        # calculate and write out the coarsened version of obs dataset to match the gcm
+        # (will be used in training)
+        coarse_obs = regrid_dataset(
+            obs_ds, gcm_one_slice, variable='tasmax', connection_string=connection_string
```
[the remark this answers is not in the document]
orianac: added to PR

cmip6_downscaling/methods/bcsd.py
orianac: I don't see an above comment? Are you recommending to shorten the docstring?

PR #31, opened 2021-11-10
Adding utilities that might be shared by multiple downscaling methods.

orianac:
Pulling out utilities from BCSD workflow that might be shared by multiple downscaling methods. Take a look here, @tcchiao and if it looks okay feel free to merge- and we have the PR here to at least show which things have been pulled out of the workflow branch (open PR #29 ).

### DIFF
--- a/cmip6_downscaling/data/cmip.py
+++ b/cmip6_downscaling/data/cmip.py
@@ -1,9 +1,10 @@
 from collections import defaultdict
-
+import os
 import intake
 import pandas as pd
 from intake_esm.merge_util import AggregationError
-
+import zarr
+import xarray as xr
 variable_ids = ['pr', 'tasmin', 'tasmax', 'rsds', 'hurs', 'ps']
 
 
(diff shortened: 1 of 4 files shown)


cmip6_downscaling/data/cmip.py
```diff
@@ -158,3 +161,74 @@ def cmip():
...
+            table_id=table_ids,
+            grid_label=grid_labels,
+            variable_id=variable_ids,
+        )
+        .df['zstore'][0]
+    )
+    if return_type == 'zarr':
+        return zarr.open_consolidated(store, mode='r')
+    elif return_type == 'xr':
+        return xr.open_zarr(store, consolidated=True)
```
tcchiao, line 218: hmm neither of these options work for me when i tried to run this with the default argument in an ipython notebook. the zarr option fails because it's missing options like account name, and the xr option fails with `.zmetadata` not found. can you specify how you're using this function?

cmip6_downscaling/data/observations.py
```diff
@@ -0,0 +1,145 @@
...
+    -------
+    [type]
+        [Chunked in space (lat=-1, lon=-1, time=1)]
+    """
+    # TEST TODO: Check that obs is chunked appropriately and throw error if not
+    # Like: assert obs.chunks == (lat=-1, lon=-1, time=1) - then eventually we can move the rechunker in as an `else`
+    regridder = xe.Regridder(obs, gcm_ds_single_time_slice, "bilinear")
+
+    obs_coarse = regridder(obs)
+    return obs_coarse
```
tcchiao, line 117: should this just be a general function called "regrid" or something along that line? how are we thinking to save the weights for downstream usage?

PR #35, opened 2021-11-12
Maca wf

tcchiao:


cmip6_downscaling/methods/detrend.py
```diff
@@ -0,0 +1,104 @@
...
+        if pstart > 0:
+            start = vector[pstart : pstart * 2]
+            vector[:pstart] = start
+        if pend > 0:
+            end = vector[-pend * 2 : -pend]
+            vector[-pend:] = end
+
+    # TODO: figure out whether use_leap_year should be true or false based on the
+    prev_year = days_in_year(data.time[0].dt.year.values - 1, use_leap_year=False)
+    next_year = days_in_year(data.time[-1].dt.year.values + 1, use_leap_year=False)
```
jhamman, line 50: You may find `DataArray.dt.is_leap_year` useful in this context: http://xarray.pydata.org/en/stable/generated/xarray.DataArray.dt.is_leap_year.html

orianac:
John thinks that the epoch adjustment window length shouldn't be a big deal (assuming we follow the same "basic linear trending approach" for the first and last 15 - or 10 - years). @tcchiao , do you have a gut instinct about whether in doing it in smaller temporal batches we'd get any strange behavior? If not (and given that John isn't concerned) perhaps we can dispense with the "dropping the beginning and ending tail years". If so, I say we leave it in.

tcchiao:
> John thinks that the epoch adjustment window length shouldn't be a big deal (assuming we follow the same "basic linear trending approach" for the first and last 15 - or 10 - years). @tcchiao , do you have a gut instinct about whether in doing it in smaller temporal batches we'd get any strange behavior? If not (and given that John isn't concerned) perhaps we can dispense with the "dropping the beginning and ending tail years". If so, I say we leave it in.

I don't have a gut instinct on whether we'll see strange behavior. The easier way for me to think about it is the percentage of our data that's "tainted" (eg if we are doing 30 year batches and we think the beginning and end 10 years are tainted, that seem a bit too much). If we don't have a strong preference let's leave the buffer code in there for now and try running with 30 year batches + 21 year rolling avg (that is, we're going to get 50 years of data for calculating trend). We can adjust if memory requirement is an issue cc @norlandrhagen @jhamman for future references.

PR #38, opened 2021-11-17
added type hinting to ERA5_transfer script

norlandrhagen:


cmip6_downscaling/workflows/ERA5/ERA5_transfer.py
```diff
@@ -16,40 +17,103 @@
...
+    fsspec.FSMap
+        fsspec mapped object
+    """
+    tgt_map = fsspec.get_mapper(tgt, connection_string)
     return tgt_map
 
 
-def copy_cleaned_data(xdf, tgt_map, overwrite=True):
-    """Copies xarray dataset to zarr store for given target"""
+def copy_cleaned_data(xdf: xr.Dataset, tgt_map: str, overwrite: bool = True):
```
jhamman, line 55: I think this could use a `-> None` return type but I'm not 100% sure. What have you seen folks use for implicit None returns?

jhamman (commented):
Just one little comment. Looks good though.

PR #41, opened 2021-11-22
Gard wf

tcchiao:


cmip6_downscaling/methods/gard.py
```diff
@@ -0,0 +1,258 @@
...
+    )
+    ss = length_scale['spatial']
+    ts = length_scale['temporal']
+
+    # reproject template into Sinusoidal projection
+    # TODO: any better ones for preserving distance between two arbitrary points on the map?
+    if 'x' not in data[label].dims:
+        template = data[label].rename({'lon': 'x', 'lat': 'y'})
+    projected = template.isel(time=0).rio.write_crs('EPSG:4326').rio.reproject(crs)
+    x = projected.x[0:100]
```
norlandrhagen: Is this a testing subset?

cmip6_downscaling/workflows/gard_flow.py
```diff
@@ -0,0 +1,209 @@
...
+    make_rechunked_obs_path,
+    make_scrf_path,
+)
+from cmip6_downscaling.workflows.utils import rechunk_zarr_array_with_caching
+
+fit_and_predict_task = task(
+    gard_fit_and_predict,
+    checkpoint=True,
+    result=XpersistResult(intermediate_cache_store, serializer=serializer),
+    target=make_gard_predict_output_path,
```
norlandrhagen, line 38: Does using params to create a path fix the caching?

cmip6_downscaling/workflows/gard_flow.py
```diff
@@ -0,0 +1,209 @@
...
+
+
+gard_postprocess_task = task(
+    gard_postprocess,
+    result=XpersistResult(results_cache_store, serializer=serializer),
+    target=make_gard_post_processed_output_path,
+)
+
+
+@task(nout=3)
```
norlandrhagen, line 56: AFAIK, if you want xpersist caching, each task can only have a single return. Could this be split into three 'prep input' tasks for caching?

norlandrhagen (commented):
Looks great! 

Let me know if I can help with debugging xpersist caching.

cmip6_downscaling/methods/gard.py
```diff
@@ -0,0 +1,258 @@
...
+        Dimension to apply the model along. Default is ``time``.
+
+    Returns
+    -------
+    output : xr.Dataset
+        GARD model prediction output. Should contain three variables: pred (predicted mean), prediction_error
+        (prediction error in fit), and exceedance_prob (probability of exceedance for threshold)
+    """
+
+    X_train = X_train.isel(lat=slice(0, 100), lon=slice(0, 100))
```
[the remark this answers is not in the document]
tcchiao: ah good catch yea it's just for testing, removed

cmip6_downscaling/methods/gard.py
```diff
@@ -0,0 +1,258 @@
...
+    temporal = gs.Gaussian(dim=1)
+    temporal.fit_variogram(bin_center, gamma, sill=np.mean(np.var(fields, axis=1)))
+
+    return {'temporal': temporal.len_scale, 'spatial': spatial.len_scale}
+
+
+def generate_scrf(
+    data: xr.Dataset,
+    label: str,
+    n_timepoints: int = 10958,
```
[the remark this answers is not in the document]
tcchiao: good catch -- the number of time points should match the prediction period length. I want to use a default value in this function but added a selection statement in the post process function to make sure they match

cmip6_downscaling/methods/gard.py
```diff
@@ -0,0 +1,258 @@
...
+        temporal_scaler=temporal_scaler,
+    )
+    ss = length_scale['spatial']
+    ts = length_scale['temporal']
+
+    # reproject template into Sinusoidal projection
+    # TODO: any better ones for preserving distance between two arbitrary points on the map?
+    if 'x' not in data[label].dims:
+        template = data[label].rename({'lon': 'x', 'lat': 'y'})
+    projected = template.isel(time=0).rio.write_crs('EPSG:4326').rio.reproject(crs)
```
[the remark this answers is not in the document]
tcchiao: `4326` is the lat/lon grid, so I don't think it would be changed. I was wondering if there's a better one to be used in place of the default for `crs` though (currently using the sinusoidal projection, see the TODO)

cmip6_downscaling/methods/gard.py
tcchiao: yes, removed

cmip6_downscaling/workflows/gard_flow.py
tcchiao: hmm i don't have a reason to believe it would help, thoughts? it would help cleaning up a lot of the params i'm passing around though

cmip6_downscaling/workflows/gard_flow.py
tcchiao: ah yup, this task is not cached separately because the rechunkers have built in caching already. we can split it later if needed but i think it's fine to leave this for now

PR #49, opened 2021-12-01
Add region segmentation code 

tcchiao:
Add two functions to:
1. determine the bounds of each subdomain according to the scientific regions definition in regionmask 
2. combine outputs from sub workflow runs on each subdomain according to regionmask


cmip6_downscaling/methods/regions.py
```diff
@@ -0,0 +1,94 @@
...
+
+    Returns
+    --------
+    ds      : the combined output where values come from the respective ds in ds_dict according to mask
+    """
+    region_codes_available = list(ds_dict.keys())
+    region_codes = np.unique(mask.values)
+    region_codes = region_codes[~np.isnan(region_codes)]
+    for code in region_codes:
+        assert code in region_codes_available
```
orianac, line 79: L75-79 feel more like a test of the fact that the region_codes were calculated correctly (is that right?) and it kind of breaks the flow of reading it all- maybe split off into a different function that's like `confirm_region_codes`? or is there another purpose for these lines?

cmip6_downscaling/methods/regions.py
```diff
@@ -0,0 +1,94 @@
...
+    out = xr.Dataset()
+    template = ds_dict[region_codes_available[0]]
+    for v in template.data_vars:
+        out[v] = xr.DataArray(
+            np.nan,
+            dims=template.dims,
+            coords={'time': template.time, 'lat': mask.lat, 'lon': mask.lon},
+        )
+        for code in region_codes:
+            downscale_output = ds_dict[code][v]
```
orianac, line 90: `downscale_output` feels like a strange name for this variable, because it isn't actually data, right? new name?

cmip6_downscaling/tests/methods/test_regions.py
```diff
@@ -0,0 +1,77 @@
+import numpy as np
+import xarray as xr
+
+from cmip6_downscaling.methods.regions import combine_outputs, generate_subdomains
+
+
+def example_3d_dataarray_us_domain(val=0):
```
orianac, line 7: oooh i love this function- will def use it!! - can we move this to somewhere else for more general use?

cmip6_downscaling/tests/methods/test_regions.py
```diff
@@ -0,0 +1,77 @@
...
+    return xr.DataArray(
+        val,
+        dims=['lat', 'lon', 'time'],
+        coords={'lat': np.arange(19, 56, 1), 'lon': np.arange(-133, -61, 2), 'time': np.arange(5)},
+    )
+
+
+def example_3d_data_us_domain():
```
orianac, line 15: and this one :)

cmip6_downscaling/tests/methods/test_regions.py
```diff
@@ -0,0 +1,77 @@
...
+
+    region_def = 'ar6'
+    buffer_size = 4
+    subdomains, mask = generate_subdomains(
+        ex_output_grid=ds.isel(time=0), buffer_size=buffer_size, region_def=region_def
+    )
+
+    # make sure the correct regions are there
+    region_codes = list(subdomains.keys())
+    assert sorted(region_codes) == [1, 2, 3, 4, 5, 6, 7, 8]
```
orianac, line 33: since region `28` is the weird one, can we have a special test for that one?

cmip6_downscaling/tests/methods/test_regions.py
```diff
@@ -0,0 +1,77 @@
...
+        for v in ds.data_vars:
+            temp[v] = xr.DataArray(k, dims=temp[v].dims, coords=temp[v].coords)
+        ds_dict[k] = temp
+
+    out = combine_outputs(ds_dict, mask)
+    for k in subdomains.keys():
+        for v in ds.data_vars:
+            unique_val = np.unique(out[v].where(mask == k).values)
+            unique_val = unique_val[~np.isnan(unique_val)]
+            assert len(unique_val) == 1
```
orianac, line 76: it's hard for me to follow what this last test is doing- maybe add a comment?

orianac (commented):
Looks good! Request for documentation, one more test, and moving some functions that are so wonderful that I want to use them in other contexts 🥇

cmip6_downscaling/methods/regions.py
```diff
@@ -0,0 +1,94 @@
...
+    ds_dict : dictionary mapping subdomain code to output
+    mask    : mask of which subdomain code to use for each grid cell (2D, lat/lon)
+
+    Returns
+    --------
+    ds      : the combined output where values come from the respective ds in ds_dict according to mask
+    """
+    region_codes_available = list(ds_dict.keys())
+    region_codes = np.unique(mask.values)
+    region_codes = region_codes[~np.isnan(region_codes)]
```
orianac, line 77: could use `np.isfinite` instead!

cmip6_downscaling/methods/regions.py
orianac: `spatially_merged_output`

orianac: disregard! but maybe just add a comment! ✨

cmip6_downscaling/methods/regions.py
```diff
@@ -0,0 +1,94 @@
...
+) -> Tuple[Dict[Union[int, float], Any], xr.DataArray]:
+    """
+    Given an example output grid, determine all subdomains that need to be process in order to generate the final output.
+    Outputs the list of bounding boxes for each subdomain considering the buffer size, as well as a mask in the resolution of the example output grid specifying
+    which subdomain's value to use for each grid cell.
+
+    Parameters
+    ----------
+    ex_output_grid : xarray dataarray or dataset. example output grid definition. both the bounding box and resolution in lat/lon directions will be used.
+    buffer_size    : int/float. for each subdomain, how much buffer area to run.
```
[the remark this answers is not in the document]
tcchiao: yes, added to the comments

cmip6_downscaling/methods/regions.py
```diff
@@ -0,0 +1,94 @@
...
+    else:
+        raise NotImplementedError('region_def must be eitehr ar6 or srex')
+
+    mask = regions.mask(ex_output_grid)
+    region_codes = np.unique(mask.values)
+    region_codes = region_codes[~np.isnan(region_codes)]
+
+    subdomains = {}
+    for n, bound in zip(regions.numbers, regions.bounds):
+        if n in region_codes:
```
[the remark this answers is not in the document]
tcchiao: this would happen if the `ex_output_grid` doesn't span the globe (e.g. only includes continental us), which in turn allows for testing the workflow with a smaller region

cmip6_downscaling/methods/regions.py
tcchiao: 💯

cmip6_downscaling/methods/regions.py
tcchiao: hmm the purpose is to ensure that the mask and ds_dict match each other, so this section of the code is similar to "validate data". I added a comment but didn't split into new function since it can't really be re-used.

cmip6_downscaling/methods/regions.py
tcchiao: renaming to single_region_output :D and added a comment

cmip6_downscaling/tests/methods/test_regions.py
tcchiao: i put it in tests/fixtures, does that work?

cmip6_downscaling/tests/methods/test_regions.py
tcchiao: good catch! the current definition for region 28 was indeed wrong. added a test and fixed the code.

cmip6_downscaling/tests/methods/test_regions.py
tcchiao: done, let me know if it doesn't help!

cmip6_downscaling/tests/methods/test_regions.py
tcchiao: done

cmip6_downscaling/methods/regions.py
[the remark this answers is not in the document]
orianac: super nit, but i think the convential format for the docstring is having the description on an indented new line, so like:
```ex_output_grid: xarray DataArray or Dataset
          Example output grid......```

File: cmip6_downscaling/methods/regions.py (hunk 9 above)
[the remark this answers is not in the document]
Comment (orianac) [reply cmip6_downscaling/methods/regions.py:43]: oh excellent.

File: cmip6_downscaling/tests/methods/test_regions.py (hunk 3 above)
Comment (orianac) [reply cmip6_downscaling/tests/methods/test_regions.py]: lovely ✨

File: cmip6_downscaling/tests/methods/test_regions.py (hunk 4 above)
Comment (orianac) [reply cmip6_downscaling/tests/methods/test_regions.py]: 🍩

File: cmip6_downscaling/tests/methods/test_regions.py (hunk 5 above)
Comment (orianac) [reply cmip6_downscaling/tests/methods/test_regions.py:19]: huzzah! 🤸‍♀️

File: cmip6_downscaling/methods/regions.py [hunk 10]
```diff
@@ -0,0 +1,99 @@
...
+    Given an example output grid, determine all subdomains that need to be process in order to generate the final output.
+    Outputs the list of bounding boxes for each subdomain considering the buffer size, as well as a mask in the resolution of the example output grid specifying
+    which subdomain's value to use for each grid cell.
+
+    Parameters
+    ----------
+    ex_output_grid : xarray dataarray or dataset. example output grid definition. both the bounding box and resolution in lat/lon directions will be used.
+    buffer_size    : int/float in unit of degree. for each subdomain, how much buffer area to run
+    region_def     : str. whether to use SREX region definition or ar6 region definition to define subdomain.
+                     see the docs https://regionmask.readthedocs.io/en/stable/defined_scientific.html for more details.
```
jhamman, line 23: Suggest reformatting to align with numpy-doc style. Something like this:

```suggestion
    ex_output_grid : xarray.DataArray or xarray.Dataset
        Example output grid definition. both the bounding box and resolution in lat/lon directions will be used.
    buffer_size : int or float
        Buffer size in unit of degree. for each subdomain, how much buffer area to run
    region_def : str
        Subregion definition name. Options are `'ar6'` or `'srex'`. See the docs https://regionmask.readthedocs.io/en/stable/defined_scientific.html for more details.
```

jhamman: It will probably be faster for both of us if I just say -- let's format all doc strings with the numpy-doc style.

cmip6_downscaling/methods/regions.py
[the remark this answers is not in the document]
tcchiao: done!

PR #54, opened 2021-12-10
Refactoring util functions

tcchiao:
Changes include:
1. a `get_gcm_grid_spec` function to get the definition of the spatial grid for a gcm 
2. change the `open_era5` function to handle multiple variables
3. new file `paths.py` to store all functions that generate paths for caching/storing 
4. a `rechunk_zarr_array_with_caching` function that handles multiple variables and caches

### DIFF
--- a/cmip6_downscaling/data/cmip.py
+++ b/cmip6_downscaling/data/cmip.py
@@ -213,7 +213,6 @@ def load_cmip(
         .df['zstore']
         .to_list()
     )
-    print(stores)
     if len(stores) > 1:
         raise ValueError('can only get 1 store at a time')
     if return_type == 'zarr':
--- a/cmip6_downscaling/data/cmip.py
+++ b/cmip6_downscaling/data/cmip.py
@@ -213,6 +213,7 @@ def load_cmip(
         .df['zstore']
         .to_list()
     )
+    print(stores)
     if len(stores) > 1:
         raise ValueError('can only get 1 store at a time')
     if return_type == 'zarr':
@@ -269,3 +270,19 @@ def gcm_munge(ds: xr.Dataset) -> xr.Dataset:
         ds = ds.drop('height')
     ds = ds.squeeze(drop=True)
     return ds
+
+
+def get_gcm_grid_spec(gcm: str) -> str:
+    gcm_grid = load_cmip(
+        source_ids=gcm,
+        return_type='xr',
+    ).sel(time=0)
+
+    nlat = len(gcm_grid.lat)
+    nlon = len(gcm_grid.lon)
+    lat_spacing = int(round(abs(gcm_grid.lat[0] - gcm_grid.lat[1]), 1) * 10)
+    lon_spacing = int(round(abs(gcm_grid.lon[0] - gcm_grid.lon[1]), 1) * 10)
+    min_lat = int(round(gcm_grid.lat.min(), 1))
+    min_lon = int(round(gcm_grid.lon.min(), 1))
+
+    return f'{nlat:d}x{nlon:d}_gridsize_{lat_spacing:d}_{lon_spacing:d}_llcorner_{min_lat:d}_{min_lon:d}'
--- a/cmip6_downscaling/data/cmip.py
+++ b/cmip6_downscaling/data/cmip.py
@@ -2,6 +2,7 @@ from collections import defaultdict
 from typing import List, Union
 
 import intake
+import numpy as np
 import pandas as pd
 import xarray as xr
 import zarr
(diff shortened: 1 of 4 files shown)


tcchiao:
@norlandrhagen @orianac a review here would be appreciated, thanks!

cmip6_downscaling/data/cmip.py
```diff
@@ -269,3 +269,19 @@ def gcm_munge(ds: xr.Dataset) -> xr.Dataset:
         ds = ds.drop('height')
     ds = ds.squeeze(drop=True)
     return ds
+
+
+def get_gcm_grid_spec(gcm: str) -> str:
+    gcm_grid = load_cmip(
```
orianac, line 276: this creates such a better string pattern than the ESMF regridder one! i went to look out of curiosity to see if there was anything we were missing and this one is way better. i wonder if we could contribute it as a more informative string-creation utility? https://github.com/pangeo-data/xESMF/blob/25bbb9a001960f46b31e5847c82216d54364ba07/xesmf/frontend.py#L356-L371

cmip6_downscaling/workflows/paths.py
```diff
@@ -0,0 +1,161 @@
...
+
+    Returns
+    -------
+    identifier : str
+        string to be used in obs related paths as specified by the params
+    """
+    if isinstance(variables, str):
+        variables = [variables]
+    var_string = '_'.join(variables)
+    return f'{obs}_{train_period_start}_{train_period_end}_{var_string}_{chunk_by}_chunked'
```
orianac: update these with preferred/clearer chunking variable- i forget which terminology we landed on as being the most intuitive. could also be `chunking_along` but i think we came up with a better solution... 🤔

cmip6_downscaling/workflows/paths.py
```diff
@@ -0,0 +1,161 @@
...
+
+
+def make_rechunked_obs_path(
```
orianac, line 38: the only difference among the three functions below is the template at the end, right? could we consolidate them into one function that just takes a template? or reads from templates defined at the top?

cmip6_downscaling/workflows/utils.py
```diff
@@ -65,10 +66,12 @@ def make_rechunker_stores(
         Stores where rechunker will write and the path to the target store
     """
     path_tmp = "az://cmip6/temp/{}.zarr".format(temp_file_name())
-    path_tgt = "az://cmip6/temp/{}.zarr".format(temp_file_name())
     temp_store = fsspec.get_mapper(path_tmp, connection_string=connection_string)
-    target_store = fsspec.get_mapper(path_tgt, connection_string=connection_string)
-    return temp_store, target_store, path_tgt
+
+    if output_path is None:
+        output_path = "az://cmip6/temp/{}.zarr".format(temp_file_name())
```
orianac, line 73: this is so much cleaner!

cmip6_downscaling/workflows/utils.py
```diff
@@ -346,3 +350,120 @@ def write_dataset(ds: xr.Dataset, path: str, chunks_dims: Tuple = ('time',)) ->
...
+    schema_dict = {}
+    for var in zarr_array.data_vars:
+        schema_dict[var] = DataArraySchema(chunks=chunks_dict[var])
+    target_schema = DatasetSchema(schema_dict)
+
+    # make storage patterns
+    temp_store, target_store, target_path = make_rechunker_stores(connection_string, output_path)
+
+    # check and see if the output is empty, if there is content, check that it's chunked correctly
+    if len(target_store) > 0:
```
orianac, line 419: when would `target_store` not be >0?

cmip6_downscaling/workflows/utils.py
```diff
@@ -346,3 +350,120 @@ def write_dataset(ds: xr.Dataset, path: str, chunks_dims: Tuple = ('time',)) ->
...
+            else:
+                raise NotImplementedError(
+                    'The content in the output path is incorrectly chunked, but overwrite is disabled.'
+                    'Either clear the output or enable overwrite by setting overwrite=True'
+                )
+
+    # process the input zarr array
+    delete_chunks_encoding(zarr_array)
+    try:
+        # now check if the input is already correctly chunked. If so, save to the output location and return
```
orianac, line 440: why would we want to do this step? this seems like an unnecessary copy (unless we want to do this just for convenient/organized data storage options)

orianac (commented):
just a few questions! and a couple suggestions but none critical

cmip6_downscaling/workflows/paths.py
tcchiao: `chunking_approach`!

cmip6_downscaling/workflows/paths.py
tcchiao: can you say more about how the templates would work?

cmip6_downscaling/workflows/utils.py
tcchiao: if the target_store is not empty

cmip6_downscaling/workflows/utils.py
tcchiao: ah, i was thinking then we don't need to rechunk when some other functions are looking for this dataset in this chunking strategy. Maybe we can add an option in the function call `cache` and only saves an extra copy if it's true?

PR #56, opened 2021-12-16
Updated funnel to xpersist -- waiting on kwargs change in PR41 of xpersist

norlandrhagen:
:building_construction: In progress :building_construction:


norlandrhagen:
- Updated funnel to xpersist.

PR #58, opened 2021-12-17
Functions for loading data and rechunking 

tcchiao:
Changes include: 
1. allow load cmip to get multiple variables at once 
2. wrapper function to combine historical/future gcm for downstream processing + rechunking 
3. allow using either a gcm name or an example grid to get gcm grid spec 
4. wrapper function to get obs and rechunk 
5. modify rechunking function to allow using an example ds for chunks and try `.chunk` if rechunker fails 
6. regrid-ds function

### DIFF
--- a/cmip6_downscaling/config/config.py
+++ b/cmip6_downscaling/config/config.py
@@ -13,7 +13,8 @@ CONNECTION_STRING = connection_string
 storage = Azure("prefect")
 intermediate_cache_path = 'az://flow-outputs/intermediate'
 intermediate_cache_store = CacheStore(intermediate_cache_path)
-results_cache_store = CacheStore('az://flow-outputs/results')
+results_cache_path = 'az://flow-outputs/results'
+results_cache_store = CacheStore(results_cache_path)
 serializer = 'xarray.zarr'
 
 # Prefect --------------------------------------------------------------------
--- a/cmip6_downscaling/config/config.py
+++ b/cmip6_downscaling/config/config.py
@@ -9,8 +9,10 @@ from prefect.storage import Azure
 
 # Azure --------------------------------------------------------------------
 connection_string = os.environ.get("AZURE_STORAGE_CONNECTION_STRING")
+CONNECTION_STRING = connection_string
 storage = Azure("prefect")
-intermediate_cache_store = CacheStore('az://flow-outputs/intermediate')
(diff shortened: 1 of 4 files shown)


cmip6_downscaling/config/config.py
```diff
@@ -9,8 +9,10 @@
 
 # Azure --------------------------------------------------------------------
 connection_string = os.environ.get("AZURE_STORAGE_CONNECTION_STRING")
+CONNECTION_STRING = connection_string
 storage = Azure("prefect")
-intermediate_cache_store = CacheStore('az://flow-outputs/intermediate')
+intermediate_cache_path = 'az://flow-outputs/intermediate'
+intermediate_cache_store = CacheStore(intermediate_cache_path)
 results_cache_store = CacheStore('az://flow-outputs/results')
```
norlandrhagen, line 16: nit: could the results cache be split into two lines to mirror the syntax of the intermediate cache path/store?

cmip6_downscaling/data/cmip.py
```diff
@@ -272,11 +285,97 @@ def gcm_munge(ds: xr.Dataset) -> xr.Dataset:
...
+        path_dict = {
+            'gcm': gcm,
+            'scenario': scenario,
+            'train_period_start': train_period_start,
+            'train_period_end': train_period_end,
+            'predict_period_start': predict_period_start,
+            'predict_period_end': predict_period_end,
+            'variables': variables,
+        }
+        rechunked_path = make_rechunked_gcm_path(chunking_approach=chunking_approach, **path_dict)
```
norlandrhagen, line 355: question. Do you think we should bundle the rechunking step in this function, or rechunk after it's returned?

tcchiao: hmm either way works i think -- bundling reduces the number of tasks in the actual flow itself, which i thought is nice because every function in the flow has to be a task. So if we were to split it out, the rechunking and make_path functions both have to be wrapped into a task. It's not difficult but feels less clean to me. Any rationale to split it out?

cmip6_downscaling/config/config.py
tcchiao: yes! done

cmip6_downscaling/data/cmip.py
norlandrhagen: Sounds good!  Less repeated tasks definitely seems clearer.

PR #60, opened 2021-12-17
Introduce common tasks to be used in multiple workflows 

tcchiao:


cmip6_downscaling/workflows/common_tasks.py
```diff
@@ -0,0 +1,271 @@
...
+    make_bias_corrected_gcm_path,
+    make_bias_corrected_obs_path,
+    make_coarse_obs_path,
+    make_interpolated_gcm_path,
+    make_interpolated_obs_path,
+)
+from cmip6_downscaling.workflows.utils import rechunk_zarr_array_with_caching, regrid_ds
+
+
+@task
```
norlandrhagen, line 31: General comment for the tasks here. Do you have any thoughts on if we should define the functions with the task decorator (and all the caching stuff) here, or in the flow. I still don't know which is better, but it would be nice to stick to one way.

cmip6_downscaling/workflows/common_tasks.py
```diff
@@ -0,0 +1,271 @@
...
+
+
+@task(
+    checkpoint=True,
+    result=XpersistResult(intermediate_cache_store, serializer=serializer),
+    target=make_coarse_obs_path,
+)
+def get_coarse_obs_task(ds_obs: xr.Dataset, gcm: str, **kwargs) -> xr.Dataset:
+    """
+    **kwargs are used to construct target file path
```
norlandrhagen, line 69: expand on docstring?

cmip6_downscaling/workflows/common_tasks.py
```diff
@@ -0,0 +1,271 @@
...
+@task(
+    checkpoint=True,
+    result=XpersistResult(intermediate_cache_store, serializer=serializer),
+    target=make_interpolated_obs_path,
+)
+def coarsen_and_interpolate_obs_task(
+    obs, train_period_start, train_period_end, variables, gcm, chunking_approach, **kwargs
+):
+    """
+    # goal here is to cache: 1) the rechunked fine obs, 2) the coarsened obs, and 3) the regridded obs
```
norlandrhagen, line 95: docstring, typehinting

norlandrhagen (commented):
Good call to move some of the tasks to a common_tasks file. Do you think we should move it out of flows/ into the tasks subdir?

cmip6_downscaling/workflows/common_tasks.py
tcchiao: hmm yea i don't have too much preference. i was sticking to the unspoken rule of only having tasks in the `workflows` (and now `tasks`) sub dir, but avoid having to repeat the task declaration in each workflow if possible. thoughts?

cmip6_downscaling/workflows/common_tasks.py
tcchiao: done

cmip6_downscaling/workflows/common_tasks.py
tcchiao: done

PR #61, opened 2021-12-20
Esmf threading

norlandrhagen:


norlandrhagen:
-Updated config for successful ESMF flows on prefect cloud.

cmip6_downscaling/methods/bcsd.py
```diff
@@ -65,54 +83,119 @@ def make_flow_paths(
...
 
 
-def return_obs(train_period_start: str, train_period_end: str, variable: str) -> xr.Dataset:
+def return_obs(
+    gcm: str,
+    scenario: str,
+    train_period_start: str,
+    train_period_end: str,
+    predict_period_start: str,
+    predict_period_end: str,
```
orianac, line 92: `predict_period` variables shouldn't be necessary in the `return_obs` step

cmip6_downscaling/methods/bcsd.py
```diff
@@ -65,54 +83,119 @@ def make_flow_paths(
...
 
 
-def get_coarse_obs(obs_ds: xr.Dataset, variable: str) -> xr.Dataset:
+def get_coarse_obs(
+    obs_ds: xr.Dataset,
+    gcm: str,
+    scenario: str,
+    train_period_start: str,
+    train_period_end: str,
+    predict_period_start: str,
```
orianac, line 146: `predict_period` variables shouldn't be required here

cmip6_downscaling/methods/bcsd.py
```diff
@@ -264,44 +378,57 @@ def return_x_train_full_time(
...
     )
-    X_train['time'] = y_full_time_ds.time.values
 
-    X_train_full_time_ds, X_train_full_time_path = rechunk_zarr_array(
-        X_train,
-        zarr_array_location=None,
-        variable=variable,
-        chunk_dims=('lat', 'lon'),
+    # Q: Ask Ori why this exists
+    gcm_train_ds_subset['time'] = coarse_obs_full_time_ds.time.values
```
orianac, line 401: this call was to force the timestamps for the cmip data to use the friendlier era5 timestamps. (i forget which dataset used which time formats). i could picture this introducing a tricky bug though (for instance if gcm timestamp didn't align for some reason) so we could use another conversion system if that is better.

orianac: or maybe just add a datetime equivalency test or something

cmip6_downscaling/methods/bcsd.py
```diff
@@ -316,63 +443,80 @@ def return_x_predict_rechunked(
...
     )
-    return x_predict_rechunked_ds
+    return gcm_predict_rechunked_ds
 
 
 def fit_and_predict(
-    x_train_rechunked_ds: xr.Dataset,
-    y_rechunked_ds: xr.Dataset,
-    x_predict_rechunked_ds: xr.Dataset,
+    gcm_train_subset_full_time_ds: xr.Dataset,  # x_train_rechunked_ds: xr.Dataset,
```
orianac, line 494: can remove the comments on the end here if we're doing away with the `x_train` and `Y_train` convention.

orianac (commented):
these changes look great! if i'm reading it right the main change of consequence is adding in the `subset_dataset` call? did i miss anything else that was big? in response, my main request is to add an `xarray_schema` chunks call for the `subset_dataset` call (this would be a way to alleviate the irregular chunks error we just talked about)

cmip6_downscaling/config/config.py
```diff
@@ -1,104 +1,203 @@
...
+        pass
+
+    @property
+    @abstractmethod
+    def executor(self) -> Any:  # pragma: no cover
+        pass
+
+
+class CloudConfig(BaseConfig):
+    def __init__(self, **kwargs):
```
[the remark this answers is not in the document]
norlandrhagen: Ah no, just some sloppy repeating of code until the merge happens.

cmip6_downscaling/methods/bcsd.py
norlandrhagen: Currently for the caching string, I have to pass all the params used for naming in. I think once we move this to Cindy's naming funcs, this can be updated.

cmip6_downscaling/workflows/utils.py
```diff
@@ -37,6 +40,47 @@ def get_store(prefix, account_key=None):
...
+    -------
+    Xarray Dataset
+        Spatially subsetted Xarray dataset.
+    """
+    subset_ds = ds.sel(
+        time=slice(start_time, end_time),
+        lon=slice(float(lonmin), float(lonmax)),
+        lat=slice(float(latmax), float(latmin)),
+    )
+    return subset_ds
```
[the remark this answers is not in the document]
norlandrhagen: Great call.

PR #65, opened 2022-01-07
add lon_to_180 util function

jhamman:


cmip6_downscaling/tests/test_utils.py
```diff
@@ -0,0 +1,43 @@
...
+    else:
+        ds = ds.assign_coords(longitude=lon, latitude=lat)
+    return ds
+
+
+@pytest.mark.parametrize(
+    "shift",
+    [
+        0,
+        -180,
```
tcchiao, line 30: when would we encounter data with a -180 shift?

cmip6_downscaling/tests/test_utils.py
```diff
@@ -0,0 +1,43 @@
...
+    ],
+)  # cant handle positive shifts yet
+def test_lon_to_180(shift):
+    xlen, ylen, zlen = (40, 20, 6)
+    ds = create_test_ds("x", "y", "lev", xlen, ylen, zlen)
+
+    ds = ds.assign_coords(x=ds.x.data + shift)
+    lon = ds["lon"].reset_coords(drop=True)
+    ds = ds.assign_coords(lon=lon + shift)
```
tcchiao, line 39: this doesn't affect the test but i'm not sure i understand why we need both x and lon in the test ds

tcchiao (approved):

PR #69, opened 2022-01-12
Convert precip to mm/day for both cmip and era5. Helps alleviate

orianac:
rounding errors when the original, tiny mm/s numbers approach zero.

Tiny PR! @tcchiao @norlandrhagen @jhamman


orianac:
@tcchiao will merging this break anything in your world?

PR #70, opened 2022-01-12
initial attempt at class data model for config

norlandrhagen:


cmip6_downscaling/config/config.py
```diff
@@ -1,104 +1,130 @@
...
+        self.agent = "az-eu-west"
+        self.storage = Azure("prefect")
+        self.extra_pip_packages = "git+https://github.com/carbonplan/cmip6-downscaling.git git+https://github.com/pangeo-data/scikit-downscale.git git+https://github.com/NCAR/xpersist.git"
+        self.intermediate_cache_path = "az://flow-outputs/intermediate"
+        self.results_cache_path = "az://flow-outputs/results"
+        self.serializer = "xarray.zarr"
+
+
+class CloudConfig(AbstractConfig):
+    def __init__(self):
```
jhamman, line 27: I think you want to decide on a semi-common set of parameters you may want to pass to each config (maybe memory, cpus, etc.). These should be configurable as keyword arguments to the class.

cmip6_downscaling/config/config.py
```diff
@@ -1,104 +1,130 @@
...
+    # prefect_executor
+    # fsspec_fs
+    def __init__(self):
+        self.connection_string = os.environ.get("AZURE_STORAGE_CONNECTION_STRING")
+        self.agent = "az-eu-west"
+        self.storage = Azure("prefect")
+        self.extra_pip_packages = "git+https://github.com/carbonplan/cmip6-downscaling.git git+https://github.com/pangeo-data/scikit-downscale.git git+https://github.com/NCAR/xpersist.git"
+        self.intermediate_cache_path = "az://flow-outputs/intermediate"
+        self.results_cache_path = "az://flow-outputs/results"
+        self.serializer = "xarray.zarr"
```
jhamman, line 23: A few things here are "generic" but most are specific to the azure cloud deployment.

cmip6_downscaling/config/config.py
```diff
@@ -1,104 +1,130 @@
...
+            "EXTRA_PIP_PACKAGES": self.extra_pip_packages,
+            "OMP_NUM_THREADS": self.OMP_NUM_THREADS,
+            "MPI_NUM_THREADS": self.MPI_NUM_THREADS,
+            "MKL_NUM_THREADS": self.MKL_NUM_THREADS,
+            "OPENBLAS_NUM_THREADS": self.OPENBLAS_NUM_THREADS,
+            "DASK_DISTRIBUTED__WORKER__RESOURCES__TASKSLOTS": self.dask_distributed_worker_resources_taskslots,
+        }
+        return env
+
+    def dask_executor(self):
```
jhamman, line 58: You probably want to rename this to simply `executor` so each config can have the same method.

cmip6_downscaling/config/config.py
```diff
@@ -1,104 +1,130 @@
...
-        adapt_kwargs={
-            "minimum": return_dask_config()["dask_executor_adapt_min"],
-            "maximum": return_dask_config()["dask_executor_adapt_max"],
-        },
-    )
-    return dask_executor
+        )
+        return daskExecutor
+
+    def kubernetes_run_config(self):
```
jhamman, line 81: rename to `run_config`

cmip6_downscaling/config/config.py
```diff
@@ -1,104 +1,130 @@
...
+def get_config(**kwargs):
+
+    if os.environ.get("CI") == "true":
+        config = LocalConfig(**kwargs)
+    elif os.environ.get("PREFECT__BACKEND") == "cloud":
+        config = CloudConfig(**kwargs)
+    elif os.environ.get("PANGEO__BACKEND") == "pangeo":
+        config = PangeoConfig(**kwargs)
+    elif os.environ.get("TEST") == "test":
+        config = TestConfig(**kwargs)
```
jhamman, line 129: needs and `else` with a `ValueError` to catch when none of the expected environment variables are set.

jhamman (changes requested):
Good start but some structural things need to change in the data model. I think the key thing to remember is that flows should be able to use any config. Imagine this workflow:

```python
config = get_config()

...

with Flow(storage=config.storage, run_config=config.run_config, executor=executor):
    ....
```` 

This flow should be runable locally, on github actions, on prefect cloud, and on our jupyterhub without any code changes. To do that, you need to make sure each config exposes exactly the same attributes/properties.

File: cmip6_downscaling/config/config.py [hunk 6]
```diff
@@ -1,104 +1,197 @@
...
+    @abstractmethod
+    def executor(self) -> Any:  # pragma: no cover
+        pass
+
+
+class CloudConfig(BaseConfig):
+    def __init__(self, **kwargs):
+        self.connection_string = os.environ.get("AZURE_STORAGE_CONNECTION_STRING")
+
+        self.intermediate_cache_path = "az://flow-outputs/intermediates"
```
orianac, line 51: is the intention to hard-code these paths as defaults? i feel like pulling that out could be useful.

cmip6_downscaling/workflows/utils.py
```diff
@@ -16,8 +15,10 @@
 
 import cmip6_downscaling.config.config as config
 
-intermediate_cache_path = config.return_azure_config()["intermediate_cache_path"]
-connection_string = config.return_azure_config()["serializer"]
+cfg = config.CloudConfig()
+
+intermediate_cache_path = cfg.intermediate_cache_path
```
orianac, line 20: how does the variable assignment for `intermediate_cache_path` differ from the `intermediate_cache_store` assignment in the flow itself?

cmip6_downscaling/workflows/utils.py
```diff
@@ -272,6 +289,9 @@ def calc_auspicious_chunks_dict(
             #  so we'll always just give it the full length of the dimension
             chunks_dict[dim] = array_dims[dim]
     # calculate the bytesize given the dtype
+    print(da)
```
orianac, line 292: probably want to delete these print statements

orianac (commented):
a couple edits/questions, but other than that looks nice!

cmip6_downscaling/workflows/utils.py
norlandrhagen: Good catch! They are identical.

cmip6_downscaling/config/config.py
norlandrhagen: Ah totally. Any thoughts on a good place for them?

orianac: maybe as a dictionary at the top that's specific to cmip6? though this definitely isn't necessary to merge - maybe just for future reference we could change it

PR #74, opened 2022-01-21
added lon_to_180 from utils.py into all loading of data in bcsd.py

norlandrhagen:


orianac (commented):
These all seem fine to me! Let's try it out!

PR #78, opened 2022-01-25
added account name to config

norlandrhagen:
- Added account_name as config var

### DIFF
--- a/cmip6_downscaling/config.py
+++ b/cmip6_downscaling/config.py
@@ -11,6 +11,7 @@ _defaults = {
     "data_catalog": {
         "cmip": "https://cmip6downscaling.blob.core.windows.net/cmip6/pangeo-cmip6.json",
         "era5": "https://cmip6downscaling.blob.core.windows.net/cmip6/ERA5_daily/",
+        "account_name": "cmip6downscaling",
     },
     "runtime": {
         "cloud": {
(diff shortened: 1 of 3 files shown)


cmip6_downscaling/config.py
```diff
@@ -11,6 +11,7 @@
     "data_catalog": {
         "cmip": "https://cmip6downscaling.blob.core.windows.net/cmip6/pangeo-cmip6.json",
         "era5": "https://cmip6downscaling.blob.core.windows.net/cmip6/ERA5_daily/",
+        "account_name": "cmip6downscaling",
```
jhamman, line 14: I would change this slightly to something like this:

```python
        "era5": {'uri': "https://cmip6downscaling.blob.core.windows.net/cmip6/ERA5_daily/", 'storage_options': {"account_name": "cmip6downscaling"}}

File: cmip6_downscaling/config.py [hunk 2]
```diff
@@ -10,8 +10,10 @@
     'methods': {'bcsd': {}, 'gard': {}, 'maca': {}},
     "data_catalog": {
         "cmip": "https://cmip6downscaling.blob.core.windows.net/cmip6/pangeo-cmip6.json",
-        "era5": "https://cmip6downscaling.blob.core.windows.net/cmip6/ERA5_daily/",
-        "account_name": "cmip6downscaling",
+        "era5": {
+            'uri': "https://cmip6downscaling.blob.core.windows.net/cmip6/ERA5_daily/",
+            'storage_options': {"account_name": "cmip6downscaling"},
+        },
```
jhamman, line 16: do the same for the `cmip` catalog to maintain the schema.

PR #79, opened 2022-01-25
get_runtime update

norlandrhagen:
changed name selection of get_runtime in runtimes.py and added new entry to config. runtime selection can be overwritten by setting env vars or overriding config options with config.set({'run_options.runtime':'local'})


cmip6_downscaling/runtimes.py
```diff
@@ -210,14 +210,14 @@ def _generate_env(self):
         return _threadsafe_env_vars
 
 
-def get_runtime(name=None, **kwargs):
-    if name == "ci":
+def get_runtime(**kwargs):
+    if config.get("run_options.runtime") == "test":
```
jhamman, line 214:
```suggestion
    if config.get("run_options.runtime") == "ci":
```

jhamman:
Made a small change but this can be merged when the tests pass.

PR #80, opened 2022-01-25
cleanup [part 1]

jhamman:
This is a long overdue repo cleanup. The goal here is to remove inactive parts of the code base and to facilitate testing of the active parts. See comments below for outstanding questions.


cmip6_downscaling/data/catalog.yaml
```diff
@@ -4,79 +4,4 @@ plugins:
...
-        description: processing method
-        type: str
-        default: "quantile-mapping"
-        allowed: ["regridded", "bias-corrected", "quantile-mapping"]
-    args:
-      urlpath: 'az://carbonplan-downscaling/cmip6/{{ method }}/conus/4000m/{{ timestep }}/{{ model }}.{{ scenario }}.{{ member }}.zarr'
-      consolidated: True
-      storage_options:
-        account_name: carbonplan
+# TODO: update with era5 and cmip6 data catalogs
```
jhamman, line 7: This is a TODO for me and/or @norlandrhagen

cmip6_downscaling/methods/bcsd.py
```diff
@@ -27,10 +24,10 @@ def make_flow_paths(
...
     PREDICT_PERIOD_END: str,
     VARIABLE: str,
-    LATMIN: str,
-    LATMAX: str,
-    LONMIN: str,
-    LONMAX: str,
+    LATMIN: str,  # Q: why is this unused
+    LATMAX: str,  # Q: why is this unused
+    LONMIN: str,  # Q: why is this unused
+    LONMAX: str,  # Q: why is this unused
```
jhamman, line 30: @orianac / @norlandrhagen - I have this question here and elsewhere. I'm probably missing something here so just explain to me what you were trying to accomplished here.

cmip6_downscaling/methods/bcsd.py
```diff
@@ -1,6 +1,3 @@
-import os
-
-os.environ["PREFECT__FLOWS__CHECKPOINTING"] = "true"
```
jhamman, line 3: I will set this on the Prefect agent so we will not need this in individual modules anymore.

cmip6_downscaling/methods/bcsd.py
norlandrhagen: I think if parameters are in the xpersist target naming, then they also have to be in the task. 

If I remove `predict_period_start` and `predict_period_end`:

```python

└── 20:31:22 | INFO    | Task 'return_obs': Starting task run...
└── 20:31:22 | ERROR   | Unexpected error: KeyError('predict_period_start')
Traceback (most recent call last):
  File "/srv/conda/envs/notebook/lib/python3.9/site-packages/prefect/engine/runner.py", line 48, in inner
    new_state = method(self, state, *args, **kwargs)
  File "/srv/conda/envs/notebook/lib/python3.9/site-packages/prefect/engine/task_runner.py", line 742, in check_target
    if result.exists(target, **formatting_kwargs):  # type: ignore
  File "/srv/conda/envs/notebook/lib/python3.9/site-packages/xpersist/prefect/result.py", line 153, in exists
    return location.format(**kwargs) in self.cache_store
KeyError: 'predict_period_start'
```

orianac: i think the goal of specifying the coordinates was to allow for running domain spatial subsets. and it is being used in the `bcsd_flow.py` script currently.

cmip6_downscaling/data/observations.py
```diff
@@ -51,12 +34,11 @@ def open_era5(variables: Union[str, List[str]], start_year: str, end_year: str)
     if isinstance(variables, str):
         variables = [variables]
 
-    stores = [
-        f'https://cmip6downscaling.blob.core.windows.net/cmip6/ERA5_daily/{y}'
-        for y in range(int(start_year), int(end_year) + 1)
-    ]
-    ds = xr.open_mfdataset(stores, engine='zarr', consolidated=True)
```
orianac, line 58: the loss of the `engine='zarr'` flag here is okay since intake handles it, right?

cmip6_downscaling/disagg/derived_variables.py
```diff
@@ -1,147 +0,0 @@
-import numpy as np
```
orianac, line 1: i could picture a world in which we want some of these (as well as the terraclimate parts) back at some point. but they aren't used now or in the near future (so deleting sounds great)

cmip6_downscaling/methods/bcsd.py
orianac: could this not be a great place to use the prefect context? so populate the `target_naming_string` template with the prefect context instead of the parameters. that way it can be passed to xpersist as an already-formatted string?

cmip6_downscaling/data/observations.py
jhamman: yep. That is encoded in the catalog now.

cmip6_downscaling/methods/bcsd.py
jhamman: I think I'm still failing to get something. The `make_flow_paths` never uses these variables. Or does it somehow and I'm just not seeing where?

orianac: ooooh!! good catch! yeah, looks like the subset bounding box never made it into that path spec! so to get this to work properly it would need to get added in lines 70-73. but this remains a situation where we might want to switch away from passing these variables around all together....

cmip6_downscaling/disagg/derived_variables.py
jhamman: totally. I can add these back if we think we're going to use them.

orianac:
Add some logging in the functions (example: log the identifiers that each task is operating on)

PR #83, opened 2022-01-26
Feature/add pyramid bcsd

norlandrhagen:


cmip6_downscaling/tasks/pyramid.py
```diff
@@ -53,8 +53,9 @@ def _postprocess(dt: dt.DataTree, levels: int, other_chunks: dict = None) -> dt.
         dt[slevel].ds = set_zarr_encoding(
             dt[slevel].ds, codec_config={"id": "zlib", "level": 1}, float_dtype="float32"
         )
-        dt[slevel].ds['time'].encoding['dtype'] = 'int32'
-        dt[slevel].ds['time_bnds'].encoding['dtype'] = 'int32'
+        for dim in ['time', 'time_bnds']:
+            if dim in dt[slevel].ds:
+                dt[slevel].ds[dim].encoding['dtype'] = 'int32'
```
jhamman, line 58:
```suggestion
        for var in ['time', 'time_bnds']:
            if var in dt[slevel].ds:
                dt[slevel].ds[var].encoding['dtype'] = 'int32'
```

cmip6_downscaling/workflows/bcsd_flow.py
```diff
@@ -14,9 +14,13 @@
     return_gcm_train_full_time,
     return_obs,
 )
+from cmip6_downscaling.tasks import pyramid
 
 runtime = runtimes.get_runtime()
 
+config.set({'storage.intermediate.uri': 'az://flow-outputs/intermediates'})
+config.set({'storage.results.uri': 'az://flow-outputs/results'})
+config.set({'storage.temporary.uri': 'az://flow-outputs/temporary'})
```
jhamman, line 23: I think this should work like this:
```suggestion
config.set(
    {
        'storage.intermediate.uri': 'az://flow-outputs/intermediates',
        'storage.results.uri': 'az://flow-outputs/results',
        'storage.temporary.uri': 'az://flow-outputs/temporary'
    }
)
```

But, before we merge this, I think we should move this sort of configuration to your `yaml` config file: https://donfig.readthedocs.io/en/latest/configuration.html#yaml-files

As a general rule, I think we should avoid setting these values in the flow module.

PR #92, opened 2022-02-01
added time.sleep for debugging esmf

norlandrhagen:


cmip6_downscaling/tasks/common_tasks.py
```diff
@@ -33,6 +33,23 @@
 get_gcm_task = task(get_gcm)
 
 
+# @task(log_stdout=True)
```
orianac, line 36: do you want this commented out? or just delete it?

cmip6_downscaling/tasks/pyramid.py
```diff
@@ -90,6 +92,7 @@ def regrid(ds: xr.Dataset, levels: int = 2, uri: str = None, other_chunks: dict
             return
 
         # create
+        time.sleep(1)
```
orianac, line 95: maybe add a comment here to explain why you're making it sleep?

cmip6_downscaling/workflows/utils.py
```diff
@@ -418,7 +419,7 @@ def regrid_dataset(
         ds_rechunked = rechunk_zarr_array_with_caching(
             zarr_array=ds, chunking_approach='full_time', max_mem='1GB'
         )
-
+    time.sleep(1)
```
orianac, line 422: copy that same "we're going to sleep here for 1 second because...." comment here and above the change below!

cmip6_downscaling/config.py
```diff
@@ -24,7 +24,7 @@
         "cloud": {
             "storage_options": {'container': 'prefect'},
             "agent": "az-eu-west",
-            "extra_pip_packages": "git+https://github.com/carbonplan/cmip6-downscaling.git git+https://github.com/pangeo-data/scikit-downscale.git",
+            "extra_pip_packages": "git+https://github.com/carbonplan/cmip6-downscaling.git@bugfix/xesmf git+https://github.com/pangeo-data/scikit-downscale.git",
```
orianac, line 27: just to be sure, you won't merge this change into main, right? i want to make sure that our cloud deployment still uses main (and that this change was only so that you could confirm that `bugfix/xesmf` works when testing)

orianac (commented):
One question and tiny comment/cleanup requests!  🚀

PR #93, opened 2022-02-02
Convert lons to 180 when loading cmip and obs 

tcchiao:


cmip6_downscaling/data/observations.py
```diff
@@ -35,6 +35,8 @@ def open_era5(variables: Union[str, List[str]], time_period: slice) -> xr.Datase
     years = range(int(time_period.start), int(time_period.stop) + 1)
 
     ds = xr.concat([cat.era5(year=year).to_dask()[variables] for year in years], dim='time')
+    ds = lon_to_180(ds)
```
tcchiao, line 38: @norlandrhagen @jhamman adding this function here (and in cmip.py) means that we're calling this function twice for the bcsd workflow. I think this is fine but want to flag

orianac: i ended up picking up this change in the analysis PR - so now it's gone but that makes sense!

cmip6_downscaling/tests/methods/test_regions.py
```diff
@@ -4,68 +4,67 @@
...
+    mask_regions = np.unique(mask.values)
+    mask_regions = mask_regions[np.isfinite(mask_regions)]
+    for r in mask_regions:
+        assert r in region_codes
+
+    # make sure the output domain is as expected
+    assert mask.lat.min().values == ds.lat.min().values
+    assert mask.lat.max().values == ds.lat.max().values
+    assert mask.lon.min().values == ds.lon.min().values
+    assert mask.lon.max().values == ds.lon.max().values
```
orianac, line 35: i assume these work great now with testing? lovely!

cmip6_downscaling/tests/methods/test_regions.py
```diff
@@ -4,68 +4,67 @@
...
+
+
+def test_combine_outputs():
+    ds = example_3d_data_us_domain()
+
+    # get mask using test data
+    region_def = 'ar6'
```
orianac, line 42: And this works for the region 28 that spans the 180/-180 boundary?

orianac: oh i'm remembering we added one specifically for this case down below. DISREGARD!

orianac (commented):
looks great- zoom zoom

PR #94, opened 2022-02-02
Cleanup util 

norlandrhagen:
Cleanup util added, if set to True, files will be removed at start of flow which should break the cache


cmip6_downscaling/tests/test_cleanup.py
```diff
@@ -0,0 +1,20 @@
...
+# from cmip6_downscaling import config
+from cmip6_downscaling.tasks.cleanup import instantiate_az_filesystem  # , remove_stores
+
+
+def test_instantiate_az_filesystem():
+    filesystem = instantiate_az_filesystem()
+    assert isinstance(filesystem, AzureBlobFileSystem)
+
+
+# Q: Should this test be included?
```
orianac, line 12: perhaps a comprehensive test of this would be to do a tiny run with the cleanup flag on and then confirm that an `fs.ls` of the bucket at the end returns a `FileNotFoundError`?

cmip6_downscaling/workflows/bcsd_flow.py
```diff
@@ -55,32 +56,42 @@
 get_coarse_obs_task = task(
     get_coarse_obs,
     tags=['dask-resource:TASKSLOTS=1'],
+    max_retries=5,
```
orianac, line 59: what do the `max_retries` and `retry_delay` arguments do? is it to give it time to get created before trying to delete it?

cmip6_downscaling/workflows/bcsd_flow.py
```diff
@@ -255,33 +267,33 @@
         uri=config.get('storage.results.uri') + pyramid_path_daily,
     )
 
-    monthly_summary_ds = monthly_summary_task(
-        postprocess_bcsd_ds,
-    )
+    # monthly_summary_ds = monthly_summary_task(
```
orianac, line 270: we should comment these back in and have it working for these as well

cmip6_downscaling/workflows/utils.py
```diff
@@ -598,9 +601,10 @@ def rechunk_zarr_array_with_caching(
                 temp_store=temp_store,
             )
             rechunk_plan.execute(retries=5)
-        rechunked_ds = xr.open_zarr(
-            target_store
-        )  # ideally we want consolidated=True but it seems that functionality isn't offered in rechunker right now
+        rechunked_ds = xr.open_zarr(target_store)
+        # remove any temp stores created by task
+        remove_stores([temp_store])
```
orianac, line 606: i assume you've tested that this works alright? like the `temp_store` doesn't end up getting used by rechunker lazily later on, right?

orianac (commented):
looks like it's getting there! this is going to be soooo useful!!! i think the one big remaining to-do is to uncomment the later steps in the flow and then get it working on the whole thing!

PR #96, opened 2022-02-03
Adding DeepSD notebooks and code 

tcchiao:


cmip6_downscaling/methods/deepsd.py
```diff
@@ -0,0 +1,136 @@
...
+
+def bilinear_interpolate(output_degree, ds):
+    target_grid_ds = xe.util.grid_global(output_degree, output_degree, cf=True)
+    regridder = xe.Regridder(ds, target_grid_ds, "bilinear", extrap_method="nearest_s2d")
+    ds_regridded = regridder(ds)
+
+    return ds_regridded
+
+
+def conservative_interpolate(output_degree, ds):
```
orianac, line 52: since this one isn't used should it be removed? is there any world in which we might want to switch to it?

orianac: oh i see it's used in the data_prep notebook. nevermind!

notebooks/deepsd_data_prep.ipynb
```diff
@@ -0,0 +1,1235 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "28209eb8-d7ce-45ba-833a-ef34c76c85b7",
+   "metadata": {},
+   "source": [
+    "## Also coarsen elevation\n",
+    "\n",
+    "Elevation data is used as auxilliary input features to the deep sd model. The\n",
```
orianac, line 375: What other input features might be useful here to include? Like, if we were to revisit this step at a later date, what other features should be top of the list to explore?

orianac (commented):
Just a couple comments. The documentation is greattttt

PR #99, opened 2022-02-04
fix to spatial anomalies and add documentation

orianac:
Tiny but mighty fix. Took opportunity to add some documentation as well. A quick scan would be great @tcchiao.


tcchiao:
looks good to me!

tcchiao (approved):

Issue #98, opened 2022-02-03
Path updates


orianac:
1. We have the same target file (make_coarse_obs_path) for both get_coarse_obs_task and return_coarse_obs_full_time_task in bcsd_flow.py. I don't think this is right, unless you're intentionally trying to overwrite? But wouldn't prefect instead just always think that the task is cached because the target already exists by the time it gets to that task? Perhaps there is some special magic though that I don't see? (Update: adding the `grid_spec` and `chunking_method` back into the `obs_identifier` should help with this)

2. Right now the pyramid files are stored with a path schema that looks like this:
    'az://flow-outputs/results_testing/pyramid_{daily,monthly,annual}/MIROC6/ssp370/tasmax/40.0_50.0_-125.0_-115.0/1981_2010/1981_2099/.pyr'. 

We settled on changing that to:  'az://flow-outputs/results_testing/pyramid_{daily,monthly,annual}/MIROC6/ssp370/bcsd/tasmax/40.0_50.0_-125.0_-115.0/1981_2010_1981_2099.zarr' So, the changes being:
* add downscale method between the ssp and the variable name
* changing the time period string into the training period and future period separated by a `_` instead of a `/` 
* Making the store be located in the time period directory instead of a subdirectory.
* 
Do these look right @jhamman?. These changes percolate not just for the pyramids but for all of the zarr stores, which should be easy because everything is housed in the `paths.py`! Woohoo!

orianac:
Also I just had a thought that fixing (1) might help with memory issues (I feel like those were cropping up at some point in the obs step @norlandrhagen? Maybe that was in a different spot so disregard if I have it confused) as the current implementation doesn't have a `full_time` chunked version of obs to work with.

PR #102, opened 2022-02-09
Path naming fixes

orianac:
@tcchiao @norlandrhagen @jhamman 

Two fixes here in response to the most critical aspects of Issue #98 (not the optional naming convention switches):

1) add the `grid_spec` back into the paths - without it we will read the same obs dataset for every GCM which wouldn't match in grid
2) add the `chunking_approach` method back in. This ensures that the coarsened obs used at any given point is chunked in the right way. This fix will make the `full_time` chunked dataset to be used for model training (it isn't being made before this PR).


tcchiao:
:+1:

PR #111, opened 2022-02-18
wip

norlandrhagen:


cmip6_downscaling/workflows/bcsd_params.json
```diff
@@ -3,6 +3,20 @@
...
+    "connection_string": "$AZURE_STORAGE_CONNECTION_STRING"
+  },
+  "config.results.storage_options": {
+    "connection_string": "$AZURE_STORAGE_CONNECTION_STRING"
+  },
+  "config.temporary.storage_options": {
+    "connection_string": "$AZURE_STORAGE_CONNECTION_STRING"
+  },
+  "runtime": "runtime",
+  "cleanup_flag": "cleanup_flag",
```
jhamman, line 19: question: can we specify a dict of config options that are arbitrarily unpacked during the flow run? This would mean just one key (`"config": {'intermediate.uri': ...}`. I would think that is a more flexible pattern.

cmip6_downscaling/workflows/bcsd_flow.py
```diff
@@ -36,22 +37,40 @@
...
+    cleanup_flag,
+):
+    config.set(
+        {
+            'config.intermediate.uri': intermediate_uri,
+            'config.results.uri': results_uri,
+            'config.temporary.uri': temporary_uri,
+            'config.intermediate.storage_options': intermediate_storage_options,
+            'config.results.storage_options': results_storage_options,
+            'config.temporary.storage_options': temporary_storage_options,
```
jhamman, line 65: I don't think there is a `config` key in the config option. Am I missing something?

cmip6_downscaling/workflows/bcsd_flow.py
```diff
@@ -144,6 +165,18 @@
...
     variable = Parameter("variable")
 
+    intermediate_uri = Parameter('config.intermediate.uri')
+    results_uri = Parameter('config.results.uri')
+    temporary_uri = Parameter('config.temporary.uri')
+    intermediate_storage_options = Parameter('config.intermediate.storage_options')
+    results_storage_options = Parameter('config.results.storage_options')
+    temporary_storage_options = Parameter('config.temporary.storage_options')
+    runtime = Parameter('runtime')
+    cleanup_flag = Parameter('cleanup_flag')
```
jhamman, line 175: see comment below about packing all of these into a single dict.

cmip6_downscaling/workflows/bcsd_flow.py
```diff
@@ -36,22 +37,40 @@
...
 
+# Transform Functions into Tasks -----------------------------------------------------------
 
-results_cache_store = CacheStore(
-    config.get("storage.results.uri"),
-    storage_options=config.get("storage.results.storage_options"),
-)
 
-# Transform Functions into Tasks -----------------------------------------------------------
+@task()
```
jhamman, line 47: If the fact that these are coming in as `Parameter`s and not raw values, we may want to try using Prefect's `context` instead.

PR #117, opened 2022-03-15
Update bcsd path

norlandrhagen:
Renamed `bcsd.py` to `core.py` and moved into `methods/bcsd/core.py`


jhamman:
closed in favor of #118

PR #119, opened 2022-03-15
Bcsd spatial anomalies task

norlandrhagen:
Updated `interpolate_obs()` task in `methods/bcsd/tasks.py`. Note: regrid is used in this task, but not yet implemented in `methods/common/utils.py`


jhamman:
closed in favor of #120

PR #120, opened 2022-03-15
regrid task

norlandrhagen:
Updated `interpolate_obs()` task in `methods/bcsd/tasks.py`. Note: regrid is used in this task, but not yet implemented in `methods/common/utils.py`


jhamman (approved):

cmip6_downscaling/methods/bcsd/tasks.py
```diff
@@ -44,13 +46,14 @@ def interpolate_obs(
...
     )
-    if use_cache and (target / '.zmetadata').exists():
-        print(f'found existing target: {target}')
+    if use_cache and (target / ".zmetadata").exists():
+        print(f"found existing target: {target}")
         return target
 
-    # TODO
+    obs_ds = xr.open_zarr(obs_path).isel(time=0)
+    interpolated_obs_ds = regrid_ds(ds_path=coarse_obs_path, target_grid_ds=obs_ds)
```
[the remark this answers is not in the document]
norlandrhagen: Took a first pass at a `regrid` tasks. Included the target_grid_path and run_parameters in the args. 

```python

def regrid(source_path: UPath, target_grid_path: UPath, run_parameters: RunParameters) -> UPath:
    ...
    return regrid_path

```

cmip6_downscaling/methods/common/tasks.py
```diff
@@ -161,6 +162,32 @@ def pyramid(
     return target
 
 
+@task
+def regrid(source_path: UPath, target_grid_path: UPath, run_parameters: RunParameters) -> UPath:
+
+    target = (
+        intermediate_dir
+        / "regrid"
+        / "_{model}_{scenario}_{variable}_{latmin}_{latmax}_{lonmin}_{lonmax}_{train_dates[0]}_{train_dates[1]}_{predict_dates[0]}_{predict_dates[1]}".format(
```
jhamman, line 171: let's create this part of the path as the combination of the `source_path` and `target_grid_path`. I don't think we need `run_parameters` to make this path.

cmip6_downscaling/methods/common/tasks.py
```diff
@@ -161,6 +162,25 @@ def pyramid(
     return target
 
 
+@task
+def regrid(source_path: UPath, target_grid_path: UPath) -> UPath:
+
+    target = intermediate_dir / "regrid" / source_path / target_grid_path
```
jhamman, line 168: I think you are going to need to do something like this: 

```suggestion
    target = intermediate_dir / "regrid" / source_path.path.replace('/', '_') + target_grid_path.path.replace('/', '_')
```

cmip6_downscaling/methods/common/tasks.py
```diff
@@ -161,6 +162,25 @@ def pyramid(
...
+
+    if use_cache and (target / '.zmetadata').exists():
+        print(f'found existing target: {target}')
+        return target
+
+    source_ds = xr.open_zarr(source_path)
+    target_grid_ds = xr.open_zarr(target_grid_path).isel(time=0)
```
jhamman, line 175: fwiw, I don't think the `isel` bit here is needed. Probably doesn't hurt but not strictly needed.

norlandrhagen: Ah good to know! I'll remove it for clarity.

cmip6_downscaling/methods/common/tasks.py
```diff
@@ -161,6 +162,25 @@ def pyramid(
     return target
 
 
+@task
```
jhamman, line 165: this will need the task slots decoration here.

jhamman:
@norlandrhagen - can you resolve the conflicts here?

PR #125, opened 2022-03-17
get_experiment task

norlandrhagen:


cmip6_downscaling/methods/bcsd/flow.py
```diff
@@ -20,6 +21,7 @@
     run_analyses,
 )
 
+dask.config.set({'temporary_directory': '/tmp/dask'})
```
jhamman, line 24: We should set this in the runtime's `__init__` method.

cmip6_downscaling/methods/common/tasks.py
```diff
@@ -55,29 +55,33 @@ def get_obs(run_parameters: RunParameters) -> UPath:
...
-    subset = subset_dataset(
-        ds, run_parameters.variable, run_parameters.train_period.time_slice, run_parameters.bbox
-    )
-    subset = subset.chunk({'time': 365})  # TODO: do better here
+    if time_subset == 'train_period':
+        time_period = run_parameters.train_period.time_slice
+    elif time_subset == 'predict_period':
+        time_period = run_parameters.predict_period.time_slice
+    else:
+        raise ValueError('time_subset arg is not train_period or predict_period.')
```
jhamman, line 81:
```suggestion
time_period = getattr(run_parameters, time_subset)
```

I think will do the same thing.

cmip6_downscaling/methods/common/tasks.py
```diff
@@ -55,29 +55,33 @@ def get_obs(run_parameters: RunParameters) -> UPath:
...
+        / time_subset
         / "{model}_{scenario}_{variable}_{latmin}_{latmax}_{lonmin}_{lonmax}_{train_dates[0]}_{train_dates[1]}_{predict_dates[0]}_{predict_dates[1]}".format(
             **asdict(run_parameters)
         )
     )
     if use_cache and (target / '.zmetadata').exists():
         print(f'found existing target: {target}')
         return target
-
+    print(target)
```
jhamman, line 71:
```suggestion
```

cmip6_downscaling/methods/common/tasks.py
```diff
@@ -55,29 +55,33 @@ def get_obs(run_parameters: RunParameters) -> UPath:
...
-def get_experiment(run_parameters: RunParameters) -> UPath:
-
-    # TODO: get train and predict data here
+def get_experiment(run_parameters: RunParameters, time_subset: str) -> UPath:
 
     target = (
         intermediate_dir
         / "get_experiment"
+        / time_subset
         / "{model}_{scenario}_{variable}_{latmin}_{latmax}_{lonmin}_{lonmax}_{train_dates[0]}_{train_dates[1]}_{predict_dates[0]}_{predict_dates[1]}".format(
```
jhamman, line 64: We should update this to only include the dates in `time_period` (set below).

cmip6_downscaling/methods/common/tasks.py
norlandrhagen: Ah this is way cleaner!

PR #142, opened 2022-03-22
get_gcm, load_cmip and get_experiment updates

norlandrhagen:


cmip6_downscaling/data/cmip.py
```diff
@@ -59,13 +53,13 @@ def postprocess(ds: xr.Dataset) -> xr.Dataset:
...
-    table_ids: str = "day",
-    grid_labels: str = "gn",
-    variable_ids: List[str] = ["tasmax"],
+    activity_ids: str,
+    experiment_ids: str,
+    member_ids: str,
+    source_ids: str,
+    table_ids: str,
+    grid_labels: str,
+    variable_ids: str,
```
jhamman, line 62:
```suggestion
    activity_ids: str = None,
    experiment_ids: str = None,
    member_ids: str = None,
    source_ids: str = None,
    table_ids: str = None,
    grid_labels: str = None,
    variable_ids: list[str] = None,
```

I think we should still be defaulting to None so these are not positional arguments. Also `variable_ids` should be a list, right?

cmip6_downscaling/data/cmip.py
```diff
@@ -93,64 +87,57 @@ def load_cmip(
...
+        col_subset = col.search(
+            activity_id=activity_ids,
+            experiment_id=experiment_ids,
+            member_id=member_ids,
+            source_id=source_ids,
+            table_id=table_ids,
+            grid_label=grid_labels,
+            variable_id=variable_ids,
+        )
+        if len(col_subset.keys()) > 1:
```
jhamman, line 100:
```suggestion
        keys = list(col_subset.keys())
        if len(keys) != 1:
```

cmip6_downscaling/data/cmip.py
```diff
@@ -93,64 +87,57 @@ def load_cmip(
...
+            raise ValueError('More than one search result exist in intak-esm search')
```
jhamman, line 101:
```suggestion
        keys = list(col_subset.keys())
        if len(keys) != 1:
            raise ValueError(f'intake-esm search returned {len(keys)}, expected exactly 1.')
```

cmip6_downscaling/data/cmip.py
```diff
@@ -93,64 +87,57 @@ def load_cmip(
...
-                .to_list()
-            )
-
-            # Q: why are we using the era5 storage_options
-            storage_options = config.get('data_catalog.era5.storage_options')
-            if len(stores) > 1:
-                raise ValueError('can only get 1 store at a time')
-
-            ds = xr.open_zarr(stores[0], storage_options=storage_options).pipe(postprocess)
+        ds = col_subset[list(col_subset.keys())[0]].to_dask().pipe(postprocess)
```
jhamman, line 103:
```suggestion
        ds = col_subset[keys[0]].to_dask().pipe(postprocess)
```

cmip6_downscaling/data/cmip.py
jhamman: if `variable_ids` is just a string, I suggest changing it to a `variable_id`.

cmip6_downscaling/data/cmip.py
```diff
@@ -93,64 +87,57 @@ def load_cmip(
...
 
-            # convert to mm/day - helpful to prevent rounding errors from very tiny numbers
-            if var == 'pr':
-                ds['pr'] *= 86400
+        # convert to mm/day - helpful to prevent rounding errors from very tiny numbers
+        if variable_ids == 'pr':
+            ds['pr'] *= 86400
```
jhamman, line 107: Another (better) option here is to check if `pr` is in `ds`:
```suggestion
        if 'pr' in ds:
            ds['pr'] *= 86400
```

Related to my comment above, this will handle the case where we get a dataset with multiple variables in it.

PR #144, opened 2022-03-22
Refactor/fullspace bugfix

orianac:
@norlandrhagen - would be great to get a review on this! thanks!! 🐈 🐈‍⬛ ❤️

### DIFF
--- a/cmip6_downscaling/methods/bcsd/tasks.py
+++ b/cmip6_downscaling/methods/bcsd/tasks.py
@@ -133,7 +133,7 @@ def fit_and_predict(
     coarse_obs_full_time_ds = xr.open_zarr(coarse_obs_full_time_path)
     experiment_train_full_time_ds = xr.open_zarr(experiment_train_full_time_path)
     experiment_predict_full_time_ds = xr.open_zarr(experiment_predict_full_time_path)
-
+    experiment_train_full_time_ds['time'] = coarse_obs_full_time_ds.time.values
     pointwise_model.fit(
         experiment_train_full_time_ds[run_parameters.variable],
         coarse_obs_full_time_ds[run_parameters.variable],
@@ -171,7 +171,7 @@ def postprocess_bcsd(
         UPath to post-processed dataset.
     """
 
-    target = results_dir / "interpolate_prediction" / run_parameters.run_id
+    target = results_dir / "daily" / run_parameters.run_id
     if use_cache and zmetadata_exists(target):
         print(f"found existing target: {target}")
         return target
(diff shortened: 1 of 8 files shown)


cmip6_downscaling/methods/bcsd/tasks.py
```diff
@@ -133,7 +133,7 @@ def fit_and_predict(
     coarse_obs_full_time_ds = xr.open_zarr(coarse_obs_full_time_path)
     experiment_train_full_time_ds = xr.open_zarr(experiment_train_full_time_path)
     experiment_predict_full_time_ds = xr.open_zarr(experiment_predict_full_time_path)
-
+    experiment_train_full_time_ds['time'] = coarse_obs_full_time_ds.time.values
```
norlandrhagen, line 136: Maybe we add a short note explaining why this is needed incase we see a similar error in the future.

cmip6_downscaling/methods/bcsd/tasks.py
```diff
@@ -171,7 +171,7 @@ def postprocess_bcsd(
         UPath to post-processed dataset.
     """
 
-    target = results_dir / "interpolate_prediction" / run_parameters.run_id
+    target = results_dir / "daily" / run_parameters.run_id
```
norlandrhagen, line 174: maybe "post-process" instead of daily?

norlandrhagen (commented):
Looks great! Just a note for future us -- at this point we were able to successfully run a full-space single year run from start to finish on a jupyter-lab instance.

orianac:
@jhamman @norlandrhagen 
As of commit [3fca939](https://github.com/carbonplan/cmip6-downscaling/pull/144/commits/3fca939ced6a50d888e1b17741b8c01eccf0486f) both a full space (1 year train, 1 year predict) and a small space (10x10 degree, 30 year train, 130 year predict) works with this setup:
- pip installing the upath fix (in the latest envs that @jhamman is installing) 
- change to kernel name in postbuild file housed in the same image fix as is getting handled in the bullet above
- rechunker v. 0.4.2 (this is NOT currently the image - we're currently using 0.3.3 - @norlandrhagen could we update that?)
currently trying out a full run!

cmip6_downscaling/methods/bcsd/tasks.py
orianac: i think that marking it `daily` denotes that it is a final output file whereas `postprocess` doesn't really mean much to me at least? the most important flag seems that it is in the `results_dir`, and from there we distinguish between timesteps. it also plays nice with the monthly/summary files that are named `monthly_summary` and `annual_summary`. but i'd be open to changing the `monthly_summary`

cmip6_downscaling/methods/bcsd/tasks.py
orianac: great point! i'll add a comment.

orianac: added a comment - @norlandrhagen - any chance you'd like to tackle this TODO?

PR #145, opened 2022-03-24
changed target paths and added cmip6_downscaling version to cache_store

norlandrhagen:


jhamman (approved):
This looks good. If this pattern persists, we may want to consolidate the path handling in a stand alone module but for now, this seems reasonably clean.

PR #146, opened 2022-03-24
time munging

norlandrhagen:
Added a time munging step to postprocess of load_cmip to shift gcm data by 12 hours to match obs.


cmip6_downscaling/data/cmip.py
```diff
@@ -47,6 +48,8 @@ def postprocess(ds: xr.Dataset) -> xr.Dataset:
     if ds.lat[0] > ds.lat[-1]:
         ds = ds.reindex({"lat": ds.lat[::-1]})
 
+    # Shifts time from Noon (12:00) start to Midnight (00:00) start to match with Obs
+    ds['time'] = ds['time'] - np.timedelta64(12, 'h')
```
jhamman, line 52: Perhaps worth trying this:

```suggestion
    ds['time'] = ds['time'].resample(time='1D').first()
```

PR #150, opened 2022-03-29
More cleanup

andersy005:
- Fixes import errors
- Adds pyupgrade pre-commit hooks 
- Bumps min Python version to 3.8 

Some modules (namely those under `workflows/`) are still not functional. It appears that some of these workflows are using tasks/functions that no longer exist. these workflows can be updated in subsequent PRs (i am currently working on the CMIP6 one).


andersy005:
pre-commit.ci autofix

andersy005:
Cc @norlandrhagen

PR #156, opened 2022-04-01
switch gard fit_and_predict to map_blocks

jhamman:


jhamman:
@orianac / @norlandrhagen - is this now in #182? Should we close this?

PR #157, opened 2022-04-04
Use dask-gateway in Pangeo runtime

jhamman:


cmip6_downscaling/runtimes.py
```diff
@@ -167,6 +168,50 @@ def _generate_env(self):
...
+            'AZURE_STORAGE_CONNECTION_STRING': os.environ['AZURE_STORAGE_CONNECTION_STRING'],
+        }
+
+        options = self._gateway.cluster_options(use_local_defaults=False)
+        options['worker_cores'] = config.get("runtime.gateway.worker_cores")
+        options['worker_memory'] = config.get("runtime.gateway.worker_memory")
+        options['image'] = config.get("runtime.gateway.image")
+        options['environment'].update(env)
+
+        self._cluster = self._gateway.new_cluster(cluster_options=options, shutdown_on_close=True)
```
jhamman, line 188: this and the `__del__` method are the last things to sort out here.

cmip6_downscaling/config.py
```diff
@@ -87,6 +87,19 @@
             "adapt_max": 20,
             "dask_distributed_worker_resources_taskslots": "1",
         },
+        "gateway": {
+            "storage_prefix": "az://",
+            "storage_options": {
+                'container': 'prefect',
+            },
+            "cluster_name": '',  #
+            "extra_pip_packages": "git+https://github.com/carbonplan/cmip6-downscaling.git@gateway",
```
norlandrhagen, line 96: Should I toss this in `envs`?

cmip6_downscaling/config.py
```diff
@@ -87,6 +87,19 @@
...
+            "image": "carbonplan/cmip6-downscaling-prefect:2022.04.03",
```
norlandrhagen, line 97: Could we share the image between the the prefect and gateway? Might be easy to update one and forget the other.

norlandrhagen (commented):
Looks great! It'll be nice to have the summary functionality.

cmip6_downscaling/config.py
jhamman: totally. we should probably pull a number of these options out to a higher level in the config.

PR #158, opened 2022-04-04
minor repo cleanup

norlandrhagen:
- removed some outdated files
- updated containers/utils str_to_hash location


norlandrhagen:
Changes in cleanup:

- Removed old cleanup bcsd paths from config.py
- removed main tag from image
- updated date to latest on image

- moved str_to_hash from containers into utils 
- updated str_to_hash import to prevent circular import 
- removed tasks/cleanup.py
- removed print statement 
- removed tests/subsets/json_params
- removed test_cleanup
- removed unused utils


Note: There are still some unused files in `workflows/`. Perhaps useful flows should be migrated out?

jhamman:
@norlandrhagen - when you get a chance, can you resolve the conflicts here and I'll give this a review

cmip6_downscaling/methods/common/containers.py
```diff
@@ -102,4 +99,4 @@ def run_id(self):
 
     @property
     def run_id_hash(self):
-        return str_to_hash(self.run_id)
+        return utils.str_to_hash(self.run_id)
```
jhamman, line 102:
```suggestion
        return str_to_hash(self.run_id)
```

cmip6_downscaling/methods/common/containers.py
```diff
@@ -1,11 +1,8 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from hashlib import blake2b
 
-
-def str_to_hash(s):
-    return blake2b(s.encode(), digest_size=8).hexdigest()
+import cmip6_downscaling.methods.common.utils as utils
```
jhamman, line 5:
```suggestion
from cmip6_downscaling.methods.common.utils import str_to_hash
```

cmip6_downscaling/methods/common/utils.py
```diff
@@ -10,7 +11,11 @@
 from xarray_schema import DataArraySchema
 from xarray_schema.base import SchemaError
 
-from cmip6_downscaling.methods.common.containers import BBox
+import cmip6_downscaling.methods.common.containers as containers
```
jhamman, line 14: I'm curious why this was needed?

jhamman (commented):
a few small comments/suggestions on imports, otherwise looks good.

cmip6_downscaling/methods/common/utils.py
norlandrhagen: There was some circular import problem where utils needed an import from containers and containers needed an import from utils.

jhamman: huh! okay, I think you can leave these as is then. Perhaps we should pull the `str_to_hash` function further up in the call stack to avoid this.

PR #163, opened 2022-04-12
Get_gcm fix

orianac:
Fixes the `get_gcm` utility such that it determines whether it should look in the historical or ssp buckets by looking at the time period of interest. Added a new argument `time_slice` to accommodate this. Previously it used the scenario argument, which works fine if you are predicting and training within the historical period. But that functionality returned an empty training period dataset when  predicting on the future.

@norlandrhagen- would be great to get your eyes on this!

### DIFF
--- a/cmip6_downscaling/data/cmip.py
+++ b/cmip6_downscaling/data/cmip.py
@@ -128,6 +128,7 @@ def get_gcm(
     grid_label: str,
     source_id: str,
     variable: str,
+    time_slice: slice
 ) -> xr.Dataset:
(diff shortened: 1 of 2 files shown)


norlandrhagen (commented):
Looks great! Should we include changes to the get_gcm calls in the flows in this PR?

orianac:
great point @norlandrhagen! thanks! this fix doesn't actually require a change in the flows, but i did add the change at the spot in the `get_experiment` task where `get_gcm` is called.

PR #169, opened 2022-04-26
config updates for prefect cloud scaling

norlandrhagen:


cmip6_downscaling/config.py
```diff
@@ -58,17 +58,17 @@
                 'container': 'prefect',
             },
             "agent": "az-eu-west",
-            "extra_pip_packages": "git+https://github.com/carbonplan/cmip6-downscaling.git",
-            "kubernetes_cpu": 7,
-            "kubernetes_memory": "16Gi",
+            "extra_pip_packages": "git+https://github.com/carbonplan/cmip6-downscaling.git git+https://github.com/pangeo-data/rechunker",
```
jhamman, line 61: do we need to update the image to include this version of rechunker?

norlandrhagen: Seems like a good idea! The version installed from source has been working great and does not have the `copy_chunk` output.

norlandrhagen:
pre-commit.ci autofix

PR #170, opened 2022-04-27
Catalog comparison notebook

norlandrhagen:
cc @orianac


orianac:
The notebook seems clear and I think it does 99% of the checking! The one comment on logic that I noticed is the line with `if len(merge_df) >= 9:`. I assume this is intending to capture if there are three ssps and three variables? I think this could break if there were (for example) four ssps but not all variables for them. This could be fixed by specifying the ssps of interest in the catalog call as `experiment_id=['ssp245', 'ssp370', 'ssp585']`. When I added that spec it seems that we have all three variables for these GCMs (some of which like CanESM5 for which we have tons of ensemble members): 

```'ACCESS-CM2', 'MPI-ESM1-2-LR', 'FGOALS-g3', 'MPI-ESM1-2-HR', 'MRI-ESM2-0', 'NorESM2-MM', 'CanESM5', 'MIROC6', 'NorESM2-LM', 'BCC-CSM2-MR', 'CMCC-ESM2', 'MIROC-ES2L', 'ACCESS-ESM1-5'```

I think the next step you suggested @norlandrhagen would be great - add a simple check to this notebook to ensure that we have the right temporal coverage for a valid ensemble member from each of these GCMs (to avoid, for example, that instance of only having 2150-2200 or whatever that curious thing was that we encountered last week @norlandrhagen). If we have at least one ensemble member that has the three ssps and three variables valid, then I think we're good to go for now and don't need to make any additional data transfers @jhamman

norlandrhagen:
Awesome, thanks for the comments @orianac!

PR #171, opened 2022-04-27
Add test notebook for merging ar6 regions

meghanrjones:
Add a notebook demonstrating merge of ar6 regions into a single xarray.Dataset.

### DIFF
--- a/notebooks/ar6_merge.ipynb
+++ b/notebooks/ar6_merge.ipynb
@@ -8,8 +8,40 @@
    "outputs": [],
    "source": [
     "import numpy as np\n",
+    "import pandas as pd\n",
     "import xarray as xr\n",
-    "import regionmask"
+    "import regionmask\n",
+    "from distributed import Client"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "98ab8398",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "client = Client()\n",
+    "client"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "ae2bf37c",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "ntime = 366  # number of time steps to load/process\n",
+    "chunk_size = 48  # x/y chunk size"
+   ]
+  },
+  {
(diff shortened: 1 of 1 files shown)


notebooks/ar6_merge.ipynb
```diff
@@ -0,0 +1,148 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "31282646-7d86-496c-a279-3bb034b144f3",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Create a new dataset with the expected shape\n",
+    "expected_ds = xr.full_like(ds[['tasmax']], np.nan)"
```
meghanrjones, line 99: I created a template dataset in this cell that is used for the final indexes because `xr.merge((da.sortby(["lat", "lon"]) for da in pieces.values()))` will not produce the same coords as the original global dataset, since there are no regions covering the southern ocean. So, there's an assumption in this that we can use an existing dataset to guide the output structure.

jhamman:
@meghanrjones - this is great and is wonderfully simple. How does this work if you remove the `.head(time=10)` from the `open_zarr` cell?

meghanrjones:
> @meghanrjones - this is great and is wonderfully simple. How does this work if you remove the `.head(time=10)` from the `open_zarr` cell?

Good question. It's currently not working on the full dataset. So far I've tried a couple alternative solutions to scale but with no success. I added this to the list for tomorrow's meeting, if you'd want to try to work it out live?

meghanrjones:
@jhamman and @norlandrhagen, I've updated the notebook with the latest revisions.

There are two methods included for merging the regions back together, "merge using map_blocks" and "merge using manual blocks". While the ``map_blocks`` method is smoother, I have found it to be unstable. It worked for a small subset of the data (``.head(time=10)``), but the "medium" server ran out of memory when merging data for the first 30 time steps. For all cases, the dask client reported large amounts of unmanaged memory. While 'merge using manual blocks' is a bit wasteful as a nested for loop, it works with less than a couple GB of memory required and ran successfully on the year long example dataset in ~10 minutes on the "medium" server.

meghanrjones:
Closing as superseded by https://github.com/carbonplan/cmip6-downscaling/pull/240

PR #172, opened 2022-04-27
use map blocks directly in bcsd fit_and_predict

jhamman:
This will hopefully help with the scaling issues in the `fit_and_predict` task.


jhamman:
pre-commit.ci autofix

PR #184, opened 2022-05-12
apply land mask to datasets in fit and predict

jhamman:
This applies the Natural Earth land mask to datasets in fit and predict. The primary motivation here is to speed up prediction.

@norlandrhagen - can you do an A/B test with this and report how this impacts the timing of a BCSD run?


cmip6_downscaling/methods/bcsd/tasks.py
```diff
@@ -174,9 +175,9 @@ def fit_and_predict(
         print(f"found existing target: {target}")
         return target
 
-    xtrain = xr.open_zarr(coarse_obs_full_time_path)
-    ytrain = xr.open_zarr(experiment_train_full_time_path)
-    xpred = xr.open_zarr(experiment_predict_full_time_path)
+    xtrain = xr.open_zarr(coarse_obs_full_time_path).pipe(apply_land_mask)
+    ytrain = xr.open_zarr(experiment_train_full_time_path).pipe(apply_land_mask)
+    xpred = xr.open_zarr(experiment_predict_full_time_path).pipe(apply_land_mask)
```
norlandrhagen, line 180:
```suggestion
    ytrain = xr.open_zarr(experiment_train_full_time_path)
    xpred = xr.open_zarr(experiment_predict_full_time_path)
```

jhamman: let's leave as is for now. I agree this is probably better but applying the mask to all input datasets is safer.

Issue #183, opened 2022-05-12
GCM lons in weights


orianac:
In the downscaling workflows we open the GCM zarr stores and adjust the lons to a `[-180,180]` range using a `postprocess` call  (https://github.com/carbonplan/cmip6-downscaling/blob/4a41f4d3c99b47127665fdeafa957255080a8825/cmip6_downscaling/data/cmip.py#L120). The weights generation flow opens GCMs directly without shifting any lons. This means that when we apply those weights they are shifted from the datasets we're applying them to. In other words, an issue arises if the dataset you're applying the weights file to has lons that differ from the dataset used to create the weights file.

I think the solution is to add the `postprocess` call to the weights generation routine -my guess is right after L67 would work (https://github.com/carbonplan/cmip6-downscaling/blob/4a41f4d3c99b47127665fdeafa957255080a8825/flows/gcm_obs_weights.py#L67). 

@andersy005 can you implement this fix? Also, might this be relevant for the pyramid generation steps since I believe they also use weights? In the event that these weights (or other weights with a similar mismatch issue) are used in other places we should check all of them. I checked the ERA5 step and it appears we use the same utility to open ERA5 in the weights generation as in the workflows (which is relevant since we adjust the `lat` ordering at https://github.com/carbonplan/cmip6-downscaling/blob/4a41f4d3c99b47127665fdeafa957255080a8825/cmip6_downscaling/data/observations.py#L45). But in case there is any other place we open ERA5 for the creation of pyramids, as long as we're using weights we need to have that same `lat` reordering implemented as well.

andersy005:
Good catch, @orianac! yes, this is relevant for the pyramid generation. 

> can you implement this fix?

i'm on it

andersy005:
I just ran into an interesting issue. our `postprocess()` function assumes that we are dealing with GCMs on regular `lat/lon` grids. However, some of the GCMs use unstructured grids, for e.g.:


<details>

<summary>MPI-M/ICON-ESM-LR</summary>

```python
In [1]: import xarray as xr

In [2]: path = 'az://cmip6/CMIP/MPI-M/ICON-ESM-LR/historical/r1i1p1f1/day/pr/gn/v20210215/'

In [3]: from cmip6_downscaling.data.cmip import postprocess

In [4]: ds = xr.open_zarr(path)

In [5]: ds
Out[5]: 
<xarray.Dataset>
[...]

PR #185, opened 2022-05-12
Shift lons and lats in pre-generated weights

andersy005:
- Towards #183


jhamman:
These changes look good to me.

PR #187, opened 2022-05-13
fix `gcm_to_obs` and `obs_to_gcm` regridding weights

andersy005:
This is  a follow-up PR to address issues that were missed by https://github.com/carbonplan/cmip6-downscaling/pull/185.

The prefect flow for re-generating the weights is currently running. will post an update once this flow is complete. 

Cc @orianac, @norlandrhagen

- Towards #183

### DIFF
--- a/cmip6_downscaling/utils.py
+++ b/cmip6_downscaling/utils.py
@@ -2,6 +2,33 @@ from __future__ import annotations
 
 from hashlib import blake2b
 
+import dask
+import datatree
+import xarray as xr
+import zarr
+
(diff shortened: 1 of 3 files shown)


andersy005:
The prefect flow took 

```python
CPU times: user 3min 42s, sys: 37.2 s, total: 4min 19s
Wall time: 43min 29s

```

to run while using the `PangeoRuntime`

@orianac & @norlandrhagen, the weights should be up-to-date. let me know if i missed anything :)

PR #188, opened 2022-05-13
Add flow for generating web catalog

andersy005:


andersy005:
pre-commit.ci autofix

PR #203, opened 2022-05-17
pass bc_kwargs to QuantileMapper

jhamman:
This fixes a soon-to-be problem where we apply detrending to `pr` (and similar variables). 

We need to add `{'detrend': False}` for `pr` and `{'detrend': True}` for `tasmax`/`tasmin` to our configs. @norlandrhagen - can you do that as part of this branch?


orianac:
@jhamman Just did another test and I think this is ready to merge and then a bump in version would be great - then to start doing a few runs.

PR #211, opened 2022-05-17
use relative imports

andersy005:
This is another attempt at re-introducing relative imports.... We'll see how this goes...


andersy005:
@norlandrhagen, i'm curious... is it safe to assume the passing test are a positive sign that everything is working fine? :)

norlandrhagen:
Hopefully haha! Maybe we can run a small test flow from this PR to see if the rel imports break it.

PR #212, opened 2022-05-19
Apply land mask

andersy005:


flows/cmip6_raw_pyramids.py
```diff
@@ -122,6 +131,20 @@ def compute_summary(assets: list[tuple(str, str)]) -> dict[str, list[str]]:
...
+    levels = len(pyramid.children)
+    for level in range(levels):
+        ds = pyramid[f'{level}'].ds
+        mask = target_pyramid[f'{level}'].ds.land
+        out = ds.where(mask > 0)
+        for variable in ds.data_vars:
+            out[variable].encoding = ds[variable].encoding
+            out[variable].attrs = ds[variable].attrs
+        pyramid[f'{level}'].ds = out
+    return pyramid
```
jhamman, line 145: can we reuse the function in `methods.common` or do we need to redefine it here for some reason?

andersy005: good catch... I had this here before putting it in `methods.common`. fixing it shortly

andersy005:
Superseded by https://github.com/carbonplan/cmip6-downscaling/pull/228

PR #220, opened 2022-06-02
Maca update

jhamman:
closes #106


cmip6_downscaling/methods/maca/core.py
```diff
@@ -251,12 +350,24 @@ def maca_construct_analogs(
...
-        lr_model.fit(xi, yi)
+
+        x_valid_mask = ~np.isnan(xi)
+        y_valid_mask = ~np.isnan(yi)
+        np.testing.assert_equal(x_valid_mask, y_valid_mask)
+
+        lr_model.fit(
+            xi[x_valid_mask],
+            yi[y_valid_mask]
+        )
```
jhamman, line 370: this is where I'm leaving off for the weekend...

The issue I was running into was that there were some nans included in the `xi` and `yi` variables and this broke the linear model fit. I was testing whether filtering nans out before bit solved the issue.

PR #221, opened 2022-06-03
fix scrf data array for temp

orianac:
For temperature make scrf a data array.


orianac:
@jhamman I'd like to merge this tiny tweak to the code path for variables not following the logistic code path- @norlandrhagen after merging, main should work for temp as well as precip now!

PR #222, opened 2022-06-03
Use land fraction to calculate more inclusive land masks

jhamman:
This is a WIP PR that expands our land mask to include coarse grid cells that have small fractions of land areas. It builds off [xesmf's tutorial](https://pangeo-xesmf.readthedocs.io/en/latest/notebooks/Masking.html) on masking. I think we need to propagate this functionality to the weights generation script (cc @andersy005).

### DIFF
--- a/cmip6_downscaling/methods/common/utils.py
+++ b/cmip6_downscaling/methods/common/utils.py
@@ -82,7 +82,50 @@ def subset_dataset(
     return subset_ds
 
 
-def apply_land_mask(ds: xr.Dataset) -> xr.Dataset:
+def _make_landfrac_mask(ds, d_fine: float = 0.1):
+
+    import xesmf as xe
+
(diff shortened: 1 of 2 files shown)


cmip6_downscaling/methods/common/utils.py
```diff
@@ -96,7 +139,7 @@ def apply_land_mask(ds: xr.Dataset) -> xr.Dataset:
     xr.Dataset
     """
 
-    mask = regionmask.defined_regions.natural_earth_v5_0_0.land_10.mask(ds)
+    mask = _add_landmask_to_dataset(ds)
```
jhamman, line 142:
```suggestion
    mask = _add_landmask_to_dataset(ds, how=how)
```

jhamman:
closed by #228

PR #224, opened 2022-06-07
Update gard and analysis from namespace to regular package

meghanrjones:
Adds ``__init__.py`` to ``cmip6_downscaling/analysis/`` and ``cmipy6_downscaling/methods/gard/``.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -22,6 +22,7 @@ jobs:
       - uses: actions/checkout@v3
       - name: Install
         run: |
+          git config --global --add safe.directory /__w/cmip6-downscaling/cmip6-downscaling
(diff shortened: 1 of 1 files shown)


meghanrjones:
I'm not sure what changed in the latest docker image to cause the new `setuptools-scm was unable to detect version` error, but I think that adding `fetch-depth: 2` to https://github.com/carbonplan/cmip6-downscaling/blob/47703f8c7f2cd1266c047234368720fd926f87b6/.github/workflows/main.yaml#L21-L22 could fix the failing CI. @jhamman do you think it would be better to investigate the changes associated with https://github.com/carbonplan/cmip6-downscaling/commit/cd66a882245d167803bd65e446bcb8790196ab55, try to update the workflow, or ignore the failure?

PR #229, opened 2022-06-15
multivariate gard implementation

orianac:
as committed here ran a test case for western US using three features to predict precip. note that this will break any caching schemes previously set up as the `run_id` now includes the feature names.

flow is functional at this point with a few remaining TODOs:

- [x] read in u/v winds from era5
- [x] read in u/v winds from cmip
- [x] read in pressure from era5
- [x] read in pressure from cmip
- [x] accommodate different transformations/detrending of different variables
- [x] attr adjustment is hardcoded inside `observations.py` - add the additional variables (and consider reading those attr spec from somewhere else or maybe moving that attr adjustment to somewhere else)
- [x] confirm that variables are passed correctly (order, shape) to sklearn
- [ ] why are there negative values in bias-corrected precip (see example in `tasks.py`)

@jhamman


norlandrhagen:
pre-commit.ci autofix

PR #238, opened 2022-06-24
Make weights_pyramid_path optional for pyramid task

meghanrjones:
This change is needed for the DeepSD flow, which produces a grid that is misaligned with the pregenerated pyramid weights.


andersy005 (approved):
👍🏽

PR #240, opened 2022-06-28
task to combine regions back together into one dataset chunked with full-time

orianac:
@jhamman 

The vast majority of this is from the ar6_regions notebook in #171 @meghanrjones. 

Anticipate finding correct specifications for the writing step. Currently implemented as dask delayed task- memory is not an issue when running 32 cores on 250GB machine. Have not tested this within the prefect set-up.

### DIFF
--- a/cmip6_downscaling/methods/maca/utils.py
+++ b/cmip6_downscaling/methods/maca/utils.py
@@ -1,8 +1,13 @@
 from __future__ import annotations
 
+import dask
 import numpy as np
 import pandas as pd
+import regionmask
 import xarray as xr
+from carbonplan_data.metadata import get_cf_global_attrs
+
+from ... import __version__ as version, config
 
 
 def add_circular_temporal_pad(data: xr.Dataset, offset: int, timeunit: str = 'D') -> xr.Dataset:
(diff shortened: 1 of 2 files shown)


cmip6_downscaling/methods/maca/utils.py
```diff
@@ -143,3 +148,104 @@ def generate_batches(
...
+    """
+    Find ar6 regions in each block, merge and reindex, write to zarr
+    """
+    components = pd.unique(mask.values.ravel())
+    components = components[components <= 45]
+    if components.size > 0:
+        merged = (
+            xr.merge(
+                xr.open_zarr(
+                    split_region_paths[ind]
```
orianac, line 236: @jhamman if antarctica regions are at the end and are dropped from the list of region paths this does not need to be changed. if not, will need to add an else statement here to ignore antarctica regions.

cmip6_downscaling/methods/maca/utils.py
```diff
@@ -143,3 +148,104 @@ def generate_batches(
...
+    return mask
+
+
+@dask.delayed()
+def merge_block_to_zarr(mask, split_region_paths, out_path, *, xslice, yslice):
+    """
+    Find ar6 regions in each block, merge and reindex, write to zarr
+    """
+    components = pd.unique(mask.values.ravel())
+    components = components[components <= 45]
```
orianac, line 231: change to integer list that is passed around in flow

cmip6_downscaling/methods/maca/utils.py
orianac: switch to `zip(integer_list, split_region_paths)`

norlandrhagen:
Closed by #106

PR #243, opened 2022-06-30
update web-catalog and cmip6-raw-pyramids flows

andersy005:


andersy005:
this is ready to be merged unless there is any objection and/or feedback

PR #261, opened 2022-08-22
initial commit prepping flows to run on aws jupyterhub

jhamman:
This PR preps our configs and catalogs to support running flows on AWS. We may want to explore a way to do this without directly editing the configs.

### DIFF
--- a/cmip6_downscaling/config.py
+++ b/cmip6_downscaling/config.py
@@ -3,43 +3,41 @@ from __future__ import annotations
 
 _defaults = {
     'auth': {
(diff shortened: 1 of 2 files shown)


jhamman:
@norlandrhagen - we had been pointing to the daily ERA5 data here: `urlpath: az://training/ERA5_daily/{{ '%04d' % year }}/`. Can you suggest an updated path? I think its fine if we continue to point at Azure for this dataset.

norlandrhagen:
> 

I believe `az://cp-cmip/training/ERA5_daily/{{ '%04d' % year }}/` is what we want.

PR #267, opened 2022-09-20
Separate DeepSD and analysis dependencies from install requirements

maxrjones:
Specify these as optional dependencies using `extras_require` (ref. https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies)


maxrjones:
/pre-commit.ci autofix

andersy005 (approved):

PR #269, opened 2022-09-21
Updating cmip6_downscaling data access

norlandrhagen:
This PR intends to update the cmip6 and era5 data retrieval methods by using intake. 

Todo:

- [x] ERA5 daily winds needs an intake catalog
- [x] L47 of Observations.py needs updating to use newly created intake catalog.


norlandrhagen:
cc @andersy005 or @orianac  would one of you mind taking a quick look at this?

PR #271, opened 2022-09-26
Update licenses

andersy005:
This PR updates licenses for both 

- MPI-ESM1-2-HR - should be CC BY 4.0 (not SA)
- MRI-ESM2-0 should be CC BY 4.0 (not SA)


andersy005:
Cc @orianac

PR #298, opened 2022-11-22
Specify theme context for monogram

maxrjones:


README.md
```diff
@@ -1,7 +1,10 @@
-<img
-  src='https://carbonplan-assets.s3.amazonaws.com/monogram/dark-small.png'
-  height='48'
-/>
+<p align="left" >
+<picture>
+  <source media="(prefers-color-scheme: dark)" srcset="https://carbonplan-assets.s3.amazonaws.com/monogram/light-small.png">
+  <source media="(prefers-color-scheme: light)" srcset="https://carbonplan-assets.s3.amazonaws.com/monogram/dark-small.png">
```
katamartin, line 4: nit: it looks like we'll get the same behavior if this is removed because we'll just fallback to the default in lightmode

katamartin (approved):

README.md
maxrjones: good point, I'll take that line out. Do you think it would be worth using this feature for the other repos as well?

katamartin: yeah I think so! maybe tracking across repos is a good candidate for the project board?

PR #308, opened 2023-01-12
update zarr store links

### LINKED ISSUE #307: Incorrect zarr store links
It turns out the `zarr store` column in https://github.com/carbonplan/cmip6-downscaling/blob/main/datasets.md contains links to zarr stores that don't exist.  For example `https://cpdataeuwest.blob.core.windows.net/cp-cmip/version1/data/DeepSD/CMIP.CCCma.CanESM5.historical.r1i1p1f1.day.DeepSD.zarr` is missing the variable in the link and as a result this points to a non-existing zarr store. To fix this issue, we should add the missing `variable` information to the links.

andersy005:
- Fixes #307


andersy005:
pre-commit.ci autofix

andersy005:
pre-commit.ci autofix

